PR #6, opened 2018-02-04
Added variable to the error creation

bradleyshawkins:
fmt.Errorf function didn't pass in the arg parameter provided. Based on the context of the log statement, I believe that is the value that should be logged. Great package! We are using it at work to send through Gmail's SMTP server.


emersion:
Thanks for your contribution!

PR #7, opened 2018-02-04
Calling User function instead of comparing User function to nil

bradleyshawkins:
The comparison to see if User was nil was comparing the function instead of the value. I added the call to the User function.


emersion:
Good catch, thanks!

Issue #8, opened 2018-03-06
Receiving email without being logged in


lupine:
Hi there,

Thanks for your work on go-imap and go-smtp! I'm attempting to use them for an email-related project, and while it seems I can use go-smtp to implement a message submission port well enough, I *can't* use it to accept email from the wider Internet (e.g., the kind of thing Exim usually does when configured as an internet site).

An SMTP transaction of this kind would generally look like:

```
EHLO example.com
MAIL FROM:...
RCPT TO:...
DATA
...
.
QUIT
```

Would you be willing to consider a PR that allows this behaviour to be optionally enabled? Something like:

```go
be := mybackend.NewSubmission()
unauthenticatedBE := mybackend.NewSMTP()

server := smtp.NewServer(backend)
server.UnauthenticatedBackend = unauthenticatedBE
```

If the `UnauthenticatedBackend` is nil, we can continue to reject as we do presently. If it's set, login is optional. If a message comes through without a login happening, the `UnauthenticatedBackend` handles it. 

We could also add:

```go
server.LoginDisabled = true
```

This would *disable* the AUTH command completely, which would make things a bit cleaner from my side :)

Happy to do the work for this, I just wanted to open a line of communication about it before starting! Let me know.

emersion:
Hi! That's a good idea indeed. Maybe we could just add a `AnonymousLogin() (User, error)` function to the `Backend` interface?

lupine:
Hmm, so we'd call it if we get to `MAIL` without an `AUTH PLAIN` being received?

The change wouldn't be backward-compatible - is that a concern?

emersion:
>Hmm, so we'd call it if we get to MAIL without an AUTH PLAIN being received?

That's the idea. We could also add `Send` to `Backend` directly, but that would duplicate the function and be confusing.

>The change wouldn't be backward-compatible - is that a concern?

No, the README explicitly states that this package is unstable. I'll consider making it stable and publishing a v1 after this change.

lupine:
Great, I'll see if I can find the time to push a branch in the next day or two for review :).

Do you agree on the `server.LoginDisabled` attribute? If I know ahead of time that I won't be accepting logins, it seems to make sense to not advertise the capability at all?

emersion:
Hmm, I wonder if it makes more sense for this to be a property of the backend. We could make `Login` optional (by moving it to another interface) for instance. Still not sure about it, just throwing ideas around.

But yes, we definitely want that.

[...]

PR #9, opened 2018-03-07
Allow go-smtp to receive anonymous emails with no option to login

lupine:
This PR implements the ideas suggested in #8:

* Allow the `AUTH *` capabilities to be disabled completely
* Allow MUAs to send emails when not logged in

The combination of these two features is ideal when running one instance of go-smtp as an MTA on port 25 with both enabled, and another instance as an MSA on port 587 with both disabled.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -64,6 +64,16 @@ func (c *Conn) init() {
 	c.text = textproto.NewConn(rwc)
 }
 
+func (c *Conn) unrecognizedCommand(cmd string) {
+	c.WriteResponse(500, fmt.Sprintf("Syntax error, %v command unrecognized", cmd))
+
+	c.nbrErrors++
+	if c.nbrErrors > 3 {
+		c.WriteResponse(500, "Too many unrecognized commands")
+		c.Close()
+	}
+}
+
 // Commands are dispatched to the appropriate handler functions.
 func (c *Conn) handle(cmd string, arg string) {
 	if cmd == "" {
@@ -94,17 +104,15 @@ func (c *Conn) handle(cmd string, arg string) {
 		c.WriteResponse(221, "Goodnight and good luck")
 		c.Close()
 	case "AUTH":
-		c.handleAuth(arg)
+		if c.server.AuthDisabled {
+			c.unrecognizedCommand(cmd)
+		} else {
+			c.handleAuth(arg)
+		}
 	case "STARTTLS":
 		c.handleStartTLS()
 	default:
-		c.WriteResponse(500, fmt.Sprintf("Syntax error, %v command unrecognized", cmd))
-
-		c.nbrErrors++
-		if c.nbrErrors > 3 {
-			c.WriteResponse(500, "Too many unrecognized commands")
-			c.Close()
-		}
+		c.unrecognizedCommand(cmd)
 	}
 }
 
@@ -138,6 +146,11 @@ func (c *Conn) IsTLS() bool {
 	return ok
 }
 
+func (c *Conn) authAllowed() bool {
+	return !c.server.AuthDisabled &&
+		(c.IsTLS() || c.server.AllowInsecureAuth)
+}
+
 // GREET state -> waiting for HELO
 func (c *Conn) handleGreet(enhanced bool, arg string) {
 	if !enhanced {
@@ -163,7 +176,7 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 		if c.server.TLSConfig != nil && !c.IsTLS() {
 			caps = append(caps, "STARTTLS")
 		}
-		if c.IsTLS() || c.server.AllowInsecureAuth {
+		if c.authAllowed() {
 			authCap := "AUTH"
 			for name, _ := range c.server.auths {
 				authCap += " " + name
--- a/conn.go
+++ b/conn.go
@@ -37,6 +37,10 @@ type Conn struct {
 	locker    sync.Mutex
 }
 
+var (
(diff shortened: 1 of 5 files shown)


conn.go
```diff
@@ -118,10 +130,12 @@ func (c *Conn) User() User {
 	return c.user
 }
 
+// Setting the user resets any message beng generated
 func (c *Conn) SetUser(user User) {
 	c.locker.Lock()
 	defer c.locker.Unlock()
 	c.user = user
+	c.msg = &message{}
```
lupine: I didn't see any case where we wanted to set the user and *not* set the message to empty. Since I was introducing a new site where the user is set, I figured I should make the dependency explicit.

conn.go
```diff
@@ -318,8 +343,6 @@ func (c *Conn) handleAuth(arg string) {
 
 	if c.User() != nil {
 		c.WriteResponse(235, "Authentication succeeded")
-
-		c.msg = &message{}
```
lupine: The sasl handlers are expected to call `SetUser`, which now sets the message too.

conn.go
```diff
@@ -421,12 +444,13 @@ func (c *Conn) ReadLine() (string, error) {
 }
 
 func (c *Conn) reset() {
-	if user := c.User(); user != nil {
```
lupine: No need to grab the lock twice.

server.go
```diff
@@ -27,6 +27,10 @@ type Server struct {
 	AllowInsecureAuth bool
 	Debug             io.Writer
 
+	// If set, the AUTH command will not be advertised and authentication
+	// attempts will be rejected. This setting overrides AllowInsecureAuth.
+	AuthDisabled bool
```
lupine: The main advantage of doing it like this, I feel, is that it mirrors the `AllowInsecureAuth` parameter. I don't feel too strongly about it though.

server_test.go
```diff
@@ -50,7 +72,15 @@ func (u *user) Logout() error {
 	return nil
 }
 
-func testServer(t *testing.T) (be *backend, s *smtp.Server, c net.Conn, scanner *bufio.Scanner) {
+type serverConfigureFunc func(*smtp.Server)
```
lupine: All this to ensure `AuthDisabled` is set to the appropriate value *before* we send `EHLO`...

conn.go
```diff
@@ -37,6 +37,10 @@ type Conn struct {
 	locker    sync.Mutex
 }
 
+var (
+	AuthRequiredErr = fmt.Errorf("Please authenticate first.")
```
emersion: The Go convention is to name error values `ErrXXX`, so the name should be `ErrAuthRequired`.

(`Backend.AnonymousLogin` doc comment should be updated too)

conn.go
emersion: Aye, that makes sense.

server_test.go
```diff
@@ -180,7 +217,7 @@ func TestServer(t *testing.T) {
 	}
 
 	if len(be.messages) != 1 {
```
emersion: Please check that the number of `anonmsgs` is zero too.

server_test.go
```diff
@@ -271,3 +324,41 @@ func TestServer_tooLongMessage(t *testing.T) {
...
+	scanner.Scan()
+	io.WriteString(c, "Hey <3\r\n")
+	io.WriteString(c, ".\r\n")
+	scanner.Scan()
+
+	if !strings.HasPrefix(scanner.Text(), "250 ") {
+		t.Fatal("Invalid DATA response:", scanner.Text())
+	}
+
+	if len(be.anonmsgs) != 1 {
```
emersion: Same: here, check that `messages` is empty.

server_test.go
```diff
@@ -180,7 +217,7 @@ func TestServer(t *testing.T) {
 	}
 
 	if len(be.messages) != 1 {
-		t.Fatal("Invalid number of sent messages:", be.messages)
+		t.Fatal("Invalid number of sent messages:", be.messages, be.anonmsgs)
```
emersion: Please check that the number of `anonmsgs` is zero too.

server.go
emersion: Yeah, in the end you're right. Using an optional interface is silly.

server_test.go
lupine: Changed in e198df1

server_test.go
lupine: Changed in e198df1

conn.go
lupine: Changed in e198df1

lupine:
Thanks @emersion , requested changes have been made.

For my use case, I only ever want one of `Login()` or `AnonymousLogin()` on my backend to function so the interface is still a little odd to me - but it's still very common for port 25 to accept both SMTP and message submission connections, so maybe it's fine for the common case :woman_shrugging: 

The introduction of anonymous receipt also increases the range of possible responses I might want to generate to the client throughout the process - temporary vs. permanent failures, etc - but I'd be happy to defer changing that until I need it, rather than coding something upfront :)

emersion:
>The introduction of anonymous receipt also increases the range of possible responses I might want to generate to the client throughout the process - temporary vs. permanent failures, etc - but I'd be happy to defer changing that until I need it, rather than coding something upfront :)

Adding this would be great! We could add some new error values when it makes sense, and maybe add an error type containing an error code if needed (just like [go-imap](https://godoc.org/github.com/emersion/go-imap/server#ErrStatusResp) does it).

PR #12, opened 2018-05-02
Add a LenientSMTP option.

### LINKED ISSUE #11: Very restrictive MAIL FROM
Most SMTP servers I've used allow `MAIL FROM: user@host`. I've never used one that required `MAIL FROM:<user@host>`.

I note that the parsing for `RCPT TO:` is far more forgiving - stripping the `<> ` chars from the trailing text.

Can I submit a PR to make this strict parsing an option?

lyda:
There's probably more to do here. Allow out of order MAIL/RCPTs and so
forth.

Fixes #11

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -8,7 +8,6 @@ import (
 	"io/ioutil"
 	"net"
 	"net/textproto"
-	"regexp"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -217,15 +217,18 @@ func (c *Conn) handleMail(arg string) {
...
 	}
 
+	mailFromRE := "(?i)^FROM:\\s*<((?:\\\\>|[^>])+|\"[^\"]+\"@[^>]+)>  ( [\\w= ]+)?$"
+	if c.server.LenientSMTP {
+		mailFromRE = "(?i)^FROM:[ <]*([^ >]*)[ >]*( [\\w=   ]+)?$"
+	}
 	// Match FROM, while accepting '>' as quoted pair and in double quoted strings
 	// (?i) makes the regex case insensitive, (?:) is non-grouping sub-match
-	re := regexp.MustCompile("(?i)^FROM:\\s*<((?:\\\\>|[^>])+|\"[^\"]+\"@[^>]+)>( [\\w= ]+)?$")
+	re := regexp.MustCompile(mailFromRE)
```
emersion: Hmmm, this regexp is a leftover from the past I really want to get rid of. Possible to parse this with `strings` instead?

server.go
```diff
@@ -25,6 +25,7 @@ type Server struct {
 	MaxIdleSeconds    int
 	MaxMessageBytes   int
 	AllowInsecureAuth bool
+	LenientSMTP       bool
```
emersion: The default behaviour should probably to be lenient (since `RCPT TO` is lenient right now). So maybe we can change this to be a "Strict" flag instead?

lyda: I didn't want to check existing behaviour, but yes, my first pass had StrictSMTP - I had to raid the antonyms page to get the most neutral opposite I could find. But happy to switch to StrictSMTP, sure.

conn.go
lyda: I honestly don't know a huge amount about the ESMTP extensions to MAIL FROM but I can go look.

emersion: Hmm, I'll just create a follow-up issue for this one if you don't feel like fixing it right now.

emersion:
Can you add a test for the new behaviour?

conn.go
lyda: Done. That removes the need for regexp everywhere.

conn.go
```diff
@@ -217,21 +216,27 @@ func (c *Conn) handleMail(arg string) {
 		c.SetUser(user)
 	}
 
-	// Match FROM, while accepting '>' as quoted pair and in double quoted strings
-	// (?i) makes the regex case insensitive, (?:) is non-grouping sub-match
-	re := regexp.MustCompile("(?i)^FROM:\\s*<((?:\\\\>|[^>])+|\"[^\"]+\"@[^>]+)>( [\\w= ]+)?$")
-	m := re.FindStringSubmatch(arg)
-	if m == nil {
+	if (len(arg) < 6) || (strings.ToUpper(arg[0:5]) != "FROM:") {
```
emersion: No need for these parentheses

conn.go
```diff
@@ -217,21 +216,27 @@ func (c *Conn) handleMail(arg string) {
...
+		c.WriteResponse(501, "Was expecting MAIL arg syntax of FROM:<address>")
+		return
+	}
+	fromArgs := strings.Split(strings.Trim(arg[5:], " "), " ")
```
emersion: What's the purpose of this `strings.Split`?

emersion: Ah, my bad, I see.

server.go
```diff
@@ -25,6 +25,7 @@ type Server struct {
 	MaxIdleSeconds    int
 	MaxMessageBytes   int
 	AllowInsecureAuth bool
+	StrictSMTP        bool
```
emersion: Since this is the `smtp` package, is it possible to rename this to simply `Strict`?

conn.go
lyda: Removed.

server.go
lyda: Renamed.

PR #27, opened 2019-01-23
added missing import of "io" in server example

lutzhorn:
The server example in `README.md` missed an import for `"io"`.


emersion:
Thanks!

PR #29, opened 2019-02-17
export SMTPError so it can be used in Send()

nberlee:
An exported smtpError struct makes it possible to have custom smtp errors codes in func Send. This is important for soft-failing (and retry later) errors.
```
func (u *User) Send(from string, to []string, r io.Reader) error {

[...]

	return &smtp.SMTPError{
		Code:    450
		Message: "Backend didn't repond, please retry later",
	}
}
```


emersion:
Thanks!

PR #30, opened 2019-03-18
Pass connection information to backend

### LINKED ISSUE #15: Add function to get connection RemoteAddr
Hi,
   it would be useful to have the remote address of the connection to implement a filter based on sender IP.
  I think it could be implemented defining User struct (UserData as User is already ad interface) with RemoteAddr field filled during init of connection.
  
  This change will impact how User is defined as at this moment the Login functions are responsible of User creation. With this change the User object is created by the connection filling all relevant data and passing it to Login functions. Is this doable?

foxcpp:
Will be honest: Didn't checked if it works correctly. But change seems to be pretty trivial so....

Closes #15.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -14,11 +14,11 @@ var (
 type Backend interface {
 	// Authenticate a user. Return smtp.ErrAuthUnsupported if you don't want to
 	// support this.
-	Login(username, password string) (User, error)
+	Login(state *ConnectionState, username, password string) (User, error)
 
 	// Called if the client attempts to send mail without logging in first.
 	// Return smtp.ErrAuthRequired if you don't want to support this.
-	AnonymousLogin() (User, error)
+	AnonymousLogin(state *ConnectionState) (User, error)
 }
 
 // An authenticated user.
--- a/conn.go
+++ b/conn.go
@@ -25,6 +25,12 @@ type message struct {
 	To []string
 }
 
+type ConnectionState struct {
+	Hostname   string
+	RemoteAddr net.Addr
+	TLS        tls.ConnectionState
+}
+
 type Conn struct {
 	conn      net.Conn
 	text      *textproto.Conn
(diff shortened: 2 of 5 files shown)


foxcpp:
I wonder when I will understand that it is a good idea to check changes more carefully before submitting PRs...

backend.go
```diff
@@ -14,11 +14,11 @@ var (
 type Backend interface {
 	// Authenticate a user. Return smtp.ErrAuthUnsupported if you don't want to
 	// support this.
-	Login(username, password string) (User, error)
+	Login(username, password string, info ConnInfo) (User, error)
```
emersion: Is it possible to make this the first argument instead? And also a pointer?

conn.go
```diff
@@ -160,6 +166,18 @@ func (c *Conn) TLSConnectionState() (state tls.ConnectionState, ok bool) {
 	return tc.ConnectionState(), true
 }
 
+func (c *Conn) Info() (info ConnInfo) {
```
emersion: Naked returns should be avoided if possible, when they don't convey a value for documentation: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters

backend.go
emersion: Okay, not making it a pointer is probably fine, seeing how `tls.ConnectionState` is passed around. Maybe rename to `smtp.ConnectionState` for consistency?

foxcpp: Sure.

conn.go
foxcpp: Fixed.

conn.go
```diff
@@ -25,6 +25,12 @@ type message struct {
 	To []string
 }
 
+type ConnInfo struct {
```
emersion: Last one and I stop bothering you: possible to rename to `ConnectionState` for consistency with the `tls` package? :P

foxcpp: Ok.

conn.go
```diff
@@ -160,6 +166,19 @@ func (c *Conn) TLSConnectionState() (state tls.ConnectionState, ok bool) {
 	return tc.ConnectionState(), true
 }
 
+func (c *Conn) Info() ConnectionState {
```
emersion: This needs to be changed as well

foxcpp: How did I... Oops.

emersion:
Thanks!

PR #31, opened 2019-03-22
Immediately pass envelope information to the backend

### LINKED ISSUE #22: Allow backend to return an error on RCPT TO or MAIL FROM
We probably need to add a new type for transactions.

foxcpp:
Closes #22.


emersion:
Thanks!

PR #33, opened 2019-03-31
Added Read and Write timeouts.

NamedKitten:
Added timeouts for Read and Write.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ func main() {
 	s.Addr = ":1025"
 	s.Domain = "localhost"
 	s.ReadTimeout = 10 * time.Second
-  s.WriteTimeout = 10 * time.Second
+	s.WriteTimeout = 10 * time.Second
 	s.MaxMessageBytes = 1024 * 1024
 	s.MaxRecipients = 50
 	s.AllowInsecureAuth = true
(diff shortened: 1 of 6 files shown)


go.sum
```diff
@@ -0,0 +1,2 @@
+github.com/emersion/go-sasl v0.0.0-20161116183048-7e096a0a6197 h1:rDJPbyliyym8ZL/Wt71kdolp6yaD4fLIQz638E6JEt0=
```
emersion: Can you remove this file? I should add it to `.gitignore`

example_test.go
```diff
@@ -139,7 +139,8 @@ func ExampleNewServer() {
 
 	s.Addr = ":1025"
 	s.Domain = "localhost"
-	s.MaxIdleSeconds = 300
+	s.WriteTimeout = 300
```
emersion: We should probably use saner values here. Something like `10 * time.Second`.

README.md
```diff
@@ -60,6 +60,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+  "time"
```
emersion: Indentation is off here

example_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"io/ioutil"
 	"log"
 	"strings"
+  "time"
```
emersion: Same

emersion:
Thanks!

PR #37, opened 2019-04-14
Server-side support for enhanced status codes

foxcpp:


conn.go
```diff
@@ -418,60 +422,64 @@ func (c *Conn) handleStartTLS() {
...
 	c.Close()
 }
 
 func (c *Conn) greet() {
-	c.WriteResponse(220, fmt.Sprintf("%v ESMTP Service Ready", c.server.Domain))
+	c.WriteResponse(220, NoEnhancedCode, fmt.Sprintf("%v ESMTP Service Ready", c.server.Domain))
 }
 
-func (c *Conn) WriteResponse(code int, text ...string) {
+func (c *Conn) WriteResponse(code int, enchCode [3]int, text ...string) {
```
emersion: `enchCode` is a weird variable name. Maybe it should be `enhCode`

foxcpp: It was a typo. Fixed.

foxcpp:
As for client support, I decided to leave it for another time because we don't return status codes in client functions at all.

conn.go
```diff
@@ -418,69 +422,87 @@ func (c *Conn) handleStartTLS() {
...
 	if c.server.WriteTimeout != 0 {
 		c.conn.SetWriteDeadline(time.Now().Add(c.server.WriteTimeout))
 	}
 
+	// All responses must include an enhanced code, if it is missing - use
+	// a generic code X.0.0.
+	if enhCode == EnhancedCodeNotSet {
+		cat := code / 100
+		switch cat {
+		case 2, 4, 5:
```
emersion: I this doesn't happen, `enhCode` will still be set to `EnhancedCodeNotSet` and will be sent as-is on the wire. I guess this isn't desirable?

data.go
```diff
@@ -6,17 +6,28 @@ import (
 
 // SMTPError specifies the error code and message that needs to be returned to the client
 type SMTPError struct {
-	Code    int
-	Message string
+	Code         int
+	EnhancedCode [3]int
```
emersion: Could we have a special `EnhancedCode` type, so that it's easier to spot constants?

foxcpp:
0.0.0 doesn't seem to be valid code, so I think we shouldn't send it

data.go
```diff
@@ -6,17 +6,28 @@ import (
...
 }
 
+// NoEnhancedCode is used to indicate that enhanced error code should not be
+// included in response.
+var NoEnhancedCode = [3]int{-1, -1, -1}
+
+// EnhancedCodeNotSet is a nil value of EnhancedCode field in SMTPError, used
+// to indicate that backend failed to provide enhanced status code. X.0.0 will
+// be used (X is derived from error code).
+var EnhancedCodeNotSet = [3]int{0, 0, 0}
```
emersion: Should this be the default for `SMTPError`? Or should `NoEnhancedCode` be the default?

foxcpp: `NoEnhancedCode` should be only used in special cases permitted by protocol (such as EHLO response). It is exported for use by extensions. 

On the other hand, RFC 2034 requires enhanced code to be included in all 2xx, 4xx and 5xx responses. I don't want to break clients expectations here because of incomplete backend implementation. Hence EnhancedCodeNotSet is used by default, kinda similar to fallback updates implementation in go-imap if backend doesn't implement the Updater interface.

conn.go
foxcpp: Oops.

foxcpp: 5be3db175beba591fd32a49ee48465fa30415448

data.go
foxcpp: bb4b508dadb2603ed121ddfaf791b082ff94fcfe

data.go
emersion: Makes sense

data.go
```diff
@@ -4,19 +4,32 @@ import (
...
 type SMTPError struct {
-	Code    int
-	Message string
+	Code         int
+	EnhancedCode EnhancedCode
+	Message      string
 }
 
+// NoEnhancedCode is used to indicate that enhanced error code should not be
+// included in response.
```
emersion: Can you clarify that this is for use by extensions, and that people should likely use `EnhancedCodeNotSet` instead?

emersion:
LGTM otherwise!

data.go
foxcpp: d37e6b0

emersion:
Thanks!

PR #42, opened 2019-05-17
Resolve #41

### LINKED ISSUE #41: Implement AUTH LOGIN method
Some thoughts on adding server support for this.

The server code already supports the PLAIN method described in https://tools.ietf.org/html/rfc4954#section-4.1
however https://tools.ietf.org/html/rfc4422#section-3 is also relevant.

Client `swaks` is liberal in using whichever PLAIN or LOGIN auth methods the server offers.
Some clients (e.g. Windows PowerShell `Send-MailMessage`) cannot use PLAIN (according to tests I've tried).

Adding a `sasl.Login` entry to `func NewServer` in `server.go` achieves most of what's needed. Then we interwork with clients such as `swaks`.
This is because `swaks` chooses to send the basic `LOGIN` command form (without username) then await the server's challenge response.

Aforementioned Windows client chooses to always send the `LOGIN base64username` form. This is not currently handled by `go-sasl/login.go` `func Next`, because it always begins in the `loginNotStarted` state, ignoring the already provided username, and proceeds to challenge for it.

https://github.com/emersion/go-sasl/blob/7e096a0a6197b89989e8cc31016daa67c8c62051/login.go#L32

Ideally we support both LOGIN formats. I'm happy to do the work. I just had a question of style for you -

a) make `go-sasl/login.go Next` be aware of the presence of an initial `response` value, assume this is the username, and go straight to `loginWaitingPassword`

or 
b) add conditional logic to https://github.com/emersion/go-smtp/blob/469c269417f0204740f34158e7e976708ea25549/conn.go#L359 to explicitly call `Next` one time before entering the `for` loop

b) limits scope of change to just this repo; but perhaps it's better to improve `Next` to handle RFC4422 sec 3 fully. 

See also
https://www.jetmore.org/john/code/swaks/latest/doc/ref.txt
https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/send-mailmessage?view=powershell-6

tuck1s:


emersion:
TBH I'm not sure we want to have this enabled by default. This LOGIN mechanism is more-or-less deprecated from what I've understood. It's easy for library users to enable LOGIN through [`Server.EnableAuth`](https://godoc.org/github.com/emersion/go-smtp#Server.EnableAuth).

tuck1s:
Ah, OK - that makes sense. I'll close this.

PR #44, opened 2019-05-21
Populate tls.Config.ServerName if empty in StartTLS

emersion:
This allows clients to just call Client.StartTLS(nil) to use the default server
name.


foxcpp:
config.Clone is added in Go 1.8.

PR #47, opened 2019-05-26
Reintroduce updated backendutil.TransformBackend after 1a1363a

mback2k:
After considering the comments in https://github.com/emersion/go-smtp/issues/22 I still think that a TransformBackend could be useful. Especially for me since it is required for my little program https://github.com/mback2k/smtp-dkim-signer.

Considering the comments in https://github.com/emersion/go-smtp/issues/22#issuecomment-475669004 I think that option 2) "Split Transform function into separate TransformFrom, TransformTo, TransformData. Pass transformed data to underlying backend instantly." makes totally sense. The mentioned aspects are:
- The underlying backend does have an ability to report errors on From/Rcpt => true
- Can't use From/To information in TransformData =>  still possible since the required state can be kept with a Session-based approach

Therefore I am proposing to bring back TransformBackend with a Session-based approach of individual Transform functions that could be bound to an object themselves. I will post an example in a separate comment. What do you think?

### DIFF
--- a/backendutil/transform.go
+++ b/backendutil/transform.go
@@ -10,10 +10,16 @@ import (
 type TransformBackend struct {
 	Backend smtp.Backend
 
-	TransformReset func()
-	TransformMail  func(from string) (string, error)
-	TransformRcpt  func(to string) (string, error)
-	TransformData  func(r io.Reader) (io.Reader, error)
+	Transform          func(username string) TransformHandler
+	AnonymousTransform func() TransformHandler
+}
+
+// TransformHandler is a container for transforming funcs.
+type TransformHandler interface {
+	TransformReset()
+	TransformMail(from string) (string, error)
+	TransformRcpt(to string) (string, error)
+	TransformData(r io.Reader) (io.Reader, error)
 }
 
 // Login implements the smtp.Backend interface.
@@ -22,7 +28,8 @@ func (be *TransformBackend) Login(state *smtp.ConnectionState, username, passwor
 	if err != nil {
 		return nil, err
 	}
-	return &transformSession{s, be}, nil
+	trans := be.Transform(username)
+	return &transformSession{s, trans}, nil
 }
 
 // AnonymousLogin implements the smtp.Backend interface.
@@ -31,22 +38,23 @@ func (be *TransformBackend) AnonymousLogin(state *smtp.ConnectionState) (smtp.Se
 	if err != nil {
 		return nil, err
 	}
-	return &transformSession{s, be}, nil
+	trans := be.AnonymousTransform()
+	return &transformSession{s, trans}, nil
 }
 
 type transformSession struct {
 	Session smtp.Session
 
-	be *TransformBackend
+	trans TransformHandler
 }
 
 func (s *transformSession) Reset() {
-	s.be.TransformReset()
+	s.trans.TransformReset()
 	s.Session.Reset()
 }
 
 func (s *transformSession) Mail(from string) error {
-	from, err := s.be.TransformMail(from)
+	from, err := s.trans.TransformMail(from)
 	if err != nil {
 		return err
 	}
@@ -54,7 +62,7 @@ func (s *transformSession) Mail(from string) error {
 }
 
 func (s *transformSession) Rcpt(to string) error {
-	to, err := s.be.TransformRcpt(to)
+	to, err := s.trans.TransformRcpt(to)
 	if err != nil {
 		return err
 	}
@@ -62,7 +70,7 @@ func (s *transformSession) Rcpt(to string) error {
 }
 
 func (s *transformSession) Data(r io.Reader) error {
-	r, err := s.be.TransformData(r)
+	r, err := s.trans.TransformData(r)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


mback2k:
Example usage can be seen here: https://github.com/mback2k/smtp-dkim-signer/blob/674e0720641237d6166dc903e37aedc180e1a495/backend.go#L58-L151

backendutil/transform.go
```diff
@@ -0,0 +1,96 @@
...
+// TransformBackend is a backend that transforms messages.
+type TransformBackend struct {
+	Backend smtp.Backend
+
+	Transform          func(session *smtp.Session, username string) TransformHandler
+	AnonymousTransform func(session *smtp.Session) TransformHandler
+}
+
+// TransformHandler is a container for transforming funcs.
+type TransformHandler interface {
```
foxcpp: This requires users to create an object that implements TransformHandler, right? Isn't it the same as implementing the actual Session interface? I guess you wanted to have a structure with func(..) variables here.

mback2k: Thanks. You are right, I probably wanted a struct instead of an interface. Will make the change later today.

mback2k:
@emersion @foxcpp I am also thinking about whether the TransformBackend should be part of this module in the package backendutil or rather a separate module, just like go-smtp-proxy. What do you think? That way it could be patched / updated / changed independently.

foxcpp:
It is relatively small (106 loc, by looking at your commits) so I don't think it should be maintained separately, it could create too much micromanagement with versions and stuff. Additionally, I think TransformBackend could often be used with simple proxy servers that do one thing (like your DKIM signer), so users could benefit from having it together with go-smtp.

mback2k:
Sounds good. The current proposal basically implements pre-hooks for the Session methods. I think it would make sense to have post-hooks to be able to handle errors as well. What do you think?

foxcpp:
Post-hooks are required so you can rollback the internal state of "transformer" when the command gets rejected by go-smtp for whatever reason, right? TBH I actually doubt the idea of having state hidden behind functions. If you need some per-session state, you probably should just implement the Session interface and use it, this will make your intent more clear. I think TransformBackend functions should be pure functions, after all, name indicates what they transform, tranforming something else would be... ehh.

mback2k:
> Post-hooks are required so you can rollback the internal state of "transformer" when the command gets rejected by go-smtp for whatever reason, right?

Yes, that is the purpose I was thinking about.

> TBH I actually doubt the idea of having state hidden behind functions. If you need some per-session state, you probably should just implement the Session interface and use it, this will make your intent more clear. I think TransformBackend functions should be pure functions, after all, name indicates what they transform, tranforming something else would be... ehh.

You are probably right. That means the TransformBackend is useful in itself for transforming purposes, but for my specific use case I will just implement the Session interface.

Do we still want to have Session-based Transform functions then? I guess it could still make sense to be able to do different transformations depending on the session / username. Otherwise we could just have the 4 (optional and nil-able) transform functions on the Backend-level again.

foxcpp:
>Otherwise we could just have the 4 (optional and nil-able) transform functions on the Backend-level again.

Yeah, I guess that's how it should be done.

emersion:
Agree with @foxcpp

mback2k:
@foxcpp @emersion Thanks for the feedback. I just incorporated all of it and pushed a new commit.

I guess for a final PR all of the individual commits can be squashed together before merge.

If you are fine with the new state, I will do that and also try to add some tests.

foxcpp:
Feel free to add more commits, we can always squash&merge it.

backendutil/transform.go
```diff
@@ -0,0 +1,81 @@
...
+	"io"
+
+	"github.com/emersion/go-smtp"
+)
+
+// TransformBackend is a backend that transforms messages.
+type TransformBackend struct {
+	Backend smtp.Backend
+
+	TransformMail func(from string) (string, error)
```
foxcpp: It would be nice to add a usage example for it, wdyt?

mback2k: I just pushed some tests for the transform backend based on the existing server tests.

Since I am not using the TransformBackend anymore, I was not able to think of a good example. For the tests I just made the transform functions encode everything using Base64.

mback2k:
I think this PR is now ready to be merged. I am looking forward to your final feedback @emersion @foxcpp or for the merge. :-)

server_test.go
```diff
@@ -405,7 +405,7 @@ func TestServer_tooManyInvalidCommands(t *testing.T) {
 }
 
 func TestServer_tooLongMessage(t *testing.T) {
```
foxcpp: Is this test relevant to TransformBackend functionality?

backendutil/transform_test.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(msg.To) != 1 || msg.To[0] != "cm9vdEBnY2hxLmdvdi51aw==" {
+		t.Fatal("Invalid mail recipients:", msg.To)
+	}
+	// base64 of "Hey <3\n" (with actual newline)
+	if string(msg.Data) != "SGV5IDwzCg==" {
+		t.Fatal("Invalid mail data:", string(msg.Data))
+	}
+}
+
+func TestServer_tooLongMessage(t *testing.T) {
```
foxcpp: Is this test relevant to TransformBackend functionality?

mback2k: Yes, it is there to check if errors are passed through correctly. I missed that part without that test initially. ;-)

foxcpp: Alright.

emersion:
Thanks!

PR #51, opened 2019-06-09
Panic handler

### LINKED ISSUE #50: Per-connection panic handler
Probably this library should have a panic handler in way similar to go-imap (emersion/go-imap#259). Wdyt?

foxcpp:
I decided to add the handler at the command level to make sure we can send a meaningful response to the connected client before dropping the connection.

Closes #50.


emersion:
Thanks!

PR #55, opened 2019-07-25
Allow empty MAIL FROM <>

mschneider82:


conn.go
```diff
@@ -263,11 +263,12 @@ func (c *Conn) handleMail(arg string) {
 			return
 		}
 	}
-	from := strings.Trim(fromArgs[0], "<> ")
+	from := strings.Trim(fromArgs[0], " ")
 	if from == "" {
 		c.WriteResponse(501, EnhancedCode{5, 5, 2}, "Was expecting MAIL arg syntax of FROM:<address>")
 		return
 	}
+	from = strings.Trim(fromArgs[0], "<>")
```
emersion: This is more strict than before: no whitespace inside angle brackets is allowed after this commit. But maybe this is a good thing?

conn.go
```diff
@@ -263,11 +263,12 @@ func (c *Conn) handleMail(arg string) {
 			return
 		}
 	}
-	from := strings.Trim(fromArgs[0], "<> ")
+	from := strings.Trim(fromArgs[0], " ")
```
emersion: No need for `strings.Trim` anymore. The string is guaranteed to be already trimmed since `fromArgs` is the result of `strings.Split`.

conn.go
mschneider82: I think it isnt more strict because Trim's 2nd parameter is a cutset. Before my change the whitespace was also stripped

emersion: I mean `MAIL FROM:      <   thing >   ` would previously be parsed as `thing`, but is now parsed as `   thing `.

conn.go
mschneider82: Good point. Will test from := fromArgs[0]
tomorrow

conn.go
mschneider82: Good point should bei same Logic noe

mschneider82:
should be OK now!

Issue #23, opened 2019-01-10
Allow backend to return per-recipient information


emersion:
e.g. success for recipient 1 and failure for recipient 2.

This is required for complete LMTP support. The client should also support this.

This could be done with a `LMTPError` implementing `error`.

mschneider82:
I have implemented this using channels to wait for rcpt status, since the rcpt order needs to be the same as incoming, you can see the _test file for a example.

mschneider82:
What do you think? I did many tests today works pretty good, also the cancel with ctx

emersion:
Fixed in https://github.com/emersion/go-smtp/pull/66

PR #56, opened 2019-07-25
Lmtp multircptstatus

mschneider82:
added support for LMTP multi rcpt status 
Also added TCP Support, works also with starttls

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ func main() {
 
 If you need more control, you can use `Client` instead.
 
-### Server
+### SMTP Server
 
 ```go
 package main
@@ -93,7 +93,7 @@ func (s *Session) Rcpt(to string) error {
 	return nil
 }
 
-func (s *Session) Data(r io.Reader) error {
+func (s *Session) Data(r io.Reader, d smtp.DataContext) error {
 	if b, err := ioutil.ReadAll(r); err != nil {
 		return err
 	} else {
@@ -140,6 +140,131 @@ DATA
 Hey <3
 .
 ```
+
+### LMTP Server
+
+```go
+package main
+
+import (
+	"context"
+	"io"
+	"io/ioutil"
+	"log"
+	"time"
+	"fmt"
+
+	"github.com/emersion/go-smtp"
+)
+
+// The Backend implements LMTP server methods.
+type Backend struct{}
+
+// Login handles a login command with username and password.
+func (be *Backend) Login(state *smtp.ConnectionState, username, password string) (smtp.Session, error) {
+	return nil, smtp.ErrAuthUnsupported
+}
+
+// AnonymousLogin requires clients to authenticate using SMTP AUTH before sending emails
+func (be *Backend) AnonymousLogin(state *smtp.ConnectionState) (smtp.Session, error) {
+	return &Session{}, nil
+}
+
+// Session is returned for every connection
+type Session struct{
+	RcptTos     []string
+}
+
+func (s *Session) Mail(from string) error {
+	log.Println("Mail from:", from)
+	return nil
+}
+
+func (s *Session) Rcpt(to string) error {
+	log.Println("Rcpt to:", to)
+	s.RcptTos = append(s.RcptTos, to)
+	return nil
+}
+
+func (s *Session) Data(r io.Reader, dataContext smtp.DataContext) error {
+	mailBytes, err := ioutil.ReadAll(r)
+	if err != nil {
+		return err
+	} 
+	
+	globalCtx := context.Background()
+
+	for i, rcpt := range s.RcptTos {
+		rcptCtx, _ := context.WithTimeout(globalCtx, 2*time.Second)
+		// we have to assing i and rcpt new to access them in the go() routine
+		rcpt := rcpt
+		i := i
+
+		dataContext.StartDelivery(rcptCtx, rcpt)
+		go func() {
+			// normaly we would deliver the mailBytes to our Maildir/HTTP backend
+			// in this case we just do a sleep 
+			time.Sleep(time.Duration(2+i) * time.Second)
+			fmt.Println(string(mailBytes))
+            // Lets finish with OK (if the request wasn't canceled because of the ctx timeout)
+			dataContext.SetStatus(rcpt, &smtp.SMTPError{
+				Code: 250,
+				EnhancedCode:  smtp.EnhancedCode{2, 0, 0},
+				Message: "Finished",
+			})
+		}()
+
+	}
+	// we always return nil in LMTP because every rcpt return code was set with dataContext.SetStatus()
+	return nil
+}
+
+func (s *Session) Reset() {
+	// we need to reset our rcptTo's slice:
+	s.RcptTos = []string{}
+}
+
+func (s *Session) Logout() error {
+	return nil
+}
+
+func main() {
+	be := &Backend{}
+
+	s := smtp.NewServer(be)
+
+	s.LMTP = true
+	s.Addr = ":1025"
+	s.Domain = "localhost"
+	s.ReadTimeout = 10 * time.Second
+	s.WriteTimeout = 10 * time.Second
+	s.MaxMessageBytes = 1024 * 1024
+	s.MaxRecipients = 50
+	s.AllowInsecureAuth = true
+
+	log.Println("Starting LMTP Server at", s.Addr)
+	if err := s.ListenAndServe(); err != nil {
+		log.Fatal(err)
+	}
+}
+```
+
+You can use the server manually with `telnet`:
+```
+$ telnet localhost 1025
+LHLO localhost
+MAIL FROM:<from@example.com>
+RCPT TO:<rcpt1@example.com>
+RCPT TO:<rcpt2@example.com>
+RCPT TO:<rcpt3@example.com>
+DATA
+Hey <3
+.
+250 2.0.0 <rcpt1@example.com> Finished
+420 4.4.7 <rcpt2@example.com> Error: timeout reached
+420 4.4.7 <rcpt3@example.com> Error: timeout reached
+```
+
 
 ## Licence
 
--- a/backend.go
+++ b/backend.go
@@ -1,6 +1,7 @@
 package smtp
 
 import (
+	"context"
 	"errors"
 	"io"
 )
@@ -33,5 +34,10 @@ type Session interface {
 	// Add recipient for currently processed message.
 	Rcpt(to string) error
 	// Set currently processed message contents and send it.
-	Data(r io.Reader) error
+	Data(r io.Reader, d DataContext) error
+}
+
+type DataContext interface {
+	SetStatus(rcpt string, status *SMTPError)
+	StartDelivery(ctx context.Context, rcpt string)
 }
--- a/conn.go
+++ b/conn.go
@@ -535,7 +535,7 @@ func (s *dataContext) SetStatus(rcpt string, status *SMTPError) {
 func (s *dataContext) StartDelivery(ctx context.Context, rcpt string) {
 	rcpt = strings.ToLower(rcpt)
 	s.rcptStatus[rcpt] = &rcptStatus{
-		ch:  make(chan *SMTPError),
+		ch:  make(chan *SMTPError, 1),
 		ctx: ctx,
 	}
 }
--- a/conn.go
+++ b/conn.go
@@ -1,6 +1,7 @@
 package smtp
 
 import (
+	"context"
 	"crypto/tls"
 	"encoding/base64"
 	"fmt"
@@ -30,14 +31,16 @@ type Conn struct {
 	session   Session
 	locker    sync.Mutex
 
-	fromReceived bool
-	recipients   []string
+	fromReceived  bool
+	recipients    []string
+	recipientsmap map[string]struct{}
 }
 
 func newConn(c net.Conn, s *Server) *Conn {
 	sc := &Conn{
-		server: s,
-		conn:   c,
+		server:        s,
+		conn:          c,
+		recipientsmap: make(map[string]struct{}),
 	}
 
 	sc.init()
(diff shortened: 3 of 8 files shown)


mschneider82:
Link to #23

mschneider82:
I found a issue when then same rcpt is used multiple times in a Mail. I will add code next Werk to tempfail If rcpt is already in []reciepients

conn.go
```diff
@@ -476,17 +495,61 @@ func (c *Conn) handleData(arg string) {
...
+	s.rcptStatus[rcpt].ch <- status
+}
+
+func (s *lmtpStatusCollector) StartDelivery(ctx context.Context, rcpt string) {
+	rcpt = strings.ToLower(rcpt)
+	s.rcptStatus[rcpt] = &rcptStatus{
+		ch:  make(chan *SMTPError),
+		ctx: ctx,
+	}
+	s.wg.Add(1)
```
emersion: I don't get why this is necessary. When `Session.Data` returns the delivery of all receipients is complete. I don't understand why it's necessary to break this assumption.

README.md
```diff
@@ -93,7 +93,7 @@ func (s *Session) Rcpt(to string) error {
 	return nil
 }
 
-func (s *Session) Data(r io.Reader) error {
+func (s *Session) Data(r io.Reader, sc smtp.LMTPStatusCollector) error {
```
emersion: Since this is only for LMTP, I wonder if there's a way to make it optional…

conn.go
mschneider82: When you are writing a lmtp server which for example stores every rcpt mails to a s3 bucket, you can do this parallel in go routines in background and once it's finished you can reply the code to the client

README.md
mschneider82: Only if we would split out lmtp completely. But its not hard for SMTPD to fill that interface, just update the data function definition

conn.go
emersion: You can just do the wait group stuff in the `Data` command.

mschneider82: The wait group is not needed anymore I will remove it

mschneider82: You need a way to tell for every rcpt the status at the end of Data. I think the easiest way to solve this it the extended interface. Even with ctx and cancel/timeout support. As lib user you don't care about the logic just use the sc interface, very easy

emersion: I don't get why `StartDelivery` is necessary at all. You just need to tell go-smtp when a recipient has succeeded/failed.

mschneider82: It's needed when you want to reply the SMTPError code immediately to the lmtpclient.

When you wait until all rcpts are finished the client may be timed out and will retry the mail. Lmtp is to prevent this, once it's delivered to a backend the client needs the status immediately

mschneider82: Please look into server_test.go how-to use it. Assume there is a http upload in the go routine or maildir delivery

emersion: >It's needed when you want to reply the SMTPError code immediately to the lmtpclient.

Then just return the error from the function.

I'd rather not change the fact that the delivery is finished exactly when the function returns. It makes everything a lot more complicated.

emersion:
LMTP replies should be in the same order as the `RCPT` commands. We should enforce this.

I'm starting to wonder whether it's worth it to go through all this trouble just to send replies as soon as possible. Maybe just a return value containing the status for each recipient is enough.

mschneider82:
The replies are correctly ordered with my code

conn.go
mschneider82: Lmtp is different here the delivery for each rcpt is a own task. If you have many rcpts and a slow delivery to the backend its important to tell tell client about the rcpt status asap. In my case I have such a problem that's why I switched to lmtp instead of smtp

emersion:
>The replies are correctly ordered since we range over recipients and ask the channels for the SMTPError in the correct order

This isn't enforced. The backend can send statuses in incorrect order, and go-smtp will mix everything up.

conn.go
emersion: Why is it important to tell clients ASAP?

I understand that it's important that the client knows which deliveries have succeeded or failed, but I'm not sure LMTP has been designed for notifying the client as soon as possible. The RFC requires recipient statuses to be sent in order, which somewhat makes that difficult.

Let's say you have 3 recipients, and the first one is very slow. You can't send the status for the completed 2 last recipients before the first one has completed.

mschneider82:
the ordering is enforced, in line 498:

https://github.com/mschneider82/go-smtp/blob/lmtp-multircptstatus/conn.go#L498

we loop over the reciepents in the correct order, once there is no status the channel read blocks until the status is available.

emersion:
>the ordering is enforced

Ah, indeed, missed that there was one channel per recipient.

conn.go
mschneider82: > Let's say you have 3 recipients, and the first one is very slow. You can't send the status for the completed 2 last recipients before the first one has completed.

Yes thats correct, the final status response is always as fast as the first rcpt delivery, this depends on the ordering in RFC, but we collect the status asap and deliver them to the client

emersion: Then I'd prefer to just return a value with the status for each recipient.

mschneider82:
Yes every rcpt has a chan for his status, or if there is a timeout or deadline the ctx will abort it and tempfails the recipient.

I am sorry that this breaks the existing api, the only way to make this would be to splitt out lmtp to a different interface and not using the smtp one. But this duplicates much code

conn.go
mschneider82: In this case the lib user has todo the ordering correctly and all the channels and context handling, then it would be better to split out lmtp to a own interface.

The other issue is that if you have 50 rcpts and a big mail the delivery can take every long, the lmtp client may times out before we finished all rcpts. In this case the lmtpclient will retry every 50 rcpt (even if there where already 48rcpts successful delivered to the backend).

README.md
foxcpp: If we are really serious about backward-compatibility - we can add an interface like that:
```go
type LMTPSession interface {
  LMTPData(r io.Reader, sc smtp.LMTPStatusCollector) error
}
```
and assert for it.

But I agree with mschneider82, it is pretty easy to fix existing code and we are at 0.x.

backend.go
```diff
@@ -33,5 +34,10 @@ type Session interface {
 	// Add recipient for currently processed message.
 	Rcpt(to string) error
 	// Set currently processed message contents and send it.
-	Data(r io.Reader) error
+	Data(r io.Reader, s LMTPStatusCollector) error
+}
+
+type LMTPStatusCollector interface {
```
foxcpp: Question. Why can't we have it as a structure? I can't think of a use case where that interface would be useful.

conn.go
```diff
@@ -476,17 +495,58 @@ func (c *Conn) handleData(arg string) {
...
+			rcptStatus := statuscollector.rcptStatus[rcpt]
+			select {
+			case <-rcptStatus.ctx.Done():
+				c.Server().ErrorLog.Printf("Context Error: %s - tempfailing", rcptStatus.ctx.Err())
+				tempfail := SMTPError{
+					Code:         420,
+					EnhancedCode: EnhancedCode{4, 4, 7},
+					Message:      "Error: timeout reached",
+				}
+				status = &tempfail
```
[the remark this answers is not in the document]
mschneider82: sure i can refactor, i just wanted to print out that its a tempfail for easier readable

conn.go
```diff
@@ -325,6 +326,23 @@ func (c *Conn) handleRcpt(arg string) {
...
+		sliceContains := func(s []string, str string) bool {
+			for _, a := range s {
+				if a == str {
+					return true
+				}
+			}
+			return false
+		}
+
+		if sliceContains(c.recipients, strings.ToLower(recipient)) {
```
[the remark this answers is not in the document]
mschneider82: a second solution would be to create a seperate map for it, i can do this

conn.go
```diff
@@ -476,17 +495,58 @@ func (c *Conn) handleData(arg string) {
...
+		rcptStatus: make(map[string]*rcptStatus),
+	}
+}
+
+func (s *lmtpStatusCollector) SetStatus(rcpt string, status *SMTPError) {
+	rcpt = strings.ToLower(rcpt)
+	s.rcptStatus[rcpt].ch <- status
+}
+
+func (s *lmtpStatusCollector) StartDelivery(ctx context.Context, rcpt string) {
```
emersion: Still not sure why calling this is necessary.

README.md
emersion: Could rename to a neutral name like `smtp.DataContext` for future extensibility.

conn.go
mschneider82: its required to have a start point to create the channel and also assign the ctx, each rcpt can have a own ctx child with seperate timeout/deadline

README.md
mschneider82: ok fine!

conn.go
emersion: Can we leave the context for later? There are too many new things in this PR.

>create the channel

We can do that when creating the `lmtpStatusCollector`

conn.go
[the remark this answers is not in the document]
mschneider82: DONE with map (see new pull req)

conn.go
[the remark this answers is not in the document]
mschneider82: DONE (see new pull req)

README.md
mschneider82: DONE with map (see new pull req)

mschneider82:
Ok everything refactored:

introduced a map for rcpt duplicate checks (better performace)
refactored do DataContext
handled the other nitpicking

server_test.go
```diff
@@ -86,6 +88,23 @@ func (s *session) Data(r io.Reader) error {
 			s.backend.messages = append(s.backend.messages, s.msg)
 		}
 	}
+
+	ctx, _ := context.WithTimeout(context.Background(), 3*time.Second)
+	for _, rcpt := range s.msg.To {
+		rcpt := rcpt
+		sc.StartDelivery(ctx, rcpt)
+		go func() {
```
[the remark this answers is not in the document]
mschneider82: once go routine per RcptData delivery seems fine to me, its the best performace, but this highly depends on the Data() implementation, the code also supports sequential processing.

conn.go
mschneider82: when creating lmtpStatusCollector/ (refactored to dataContext), we dont have a ctx per rcpt. Without a ctx we cannot implement a timeout, but this is highly needed

emersion: 1. Isn't one context for all recipients enough?
2. Why do we need to add this to go-smtp? Can't backends implement a timeout internally?
3. As said above, I'm not a fan of "big pull requests" adding a lot of new features.

mschneider82: currently we dont have a context support in the library, There should be a way to get the ctx from the library user ( i think StartDelivery is a elegant way, you can use the same ctx as in a http library ), its not forced to use a ctx for each rcpt, but its possible

it isnt that big pull request, I rebased it into one commit, we just add the feature "multi rcpt status", of couse with ctx/channel support, if a user doesnt want to parallel delivery to his backend he just have to not use go() routines in his code. we collected the status just with channels thats the go way.

mschneider82:
I have added README example code for lmtp

server_test.go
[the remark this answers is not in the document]
foxcpp: Alright.

server.go
```diff
@@ -147,15 +146,15 @@ func (s *Server) handleConn(c *Conn) error {
 // ListenAndServe listens on the network address s.Addr and then calls Serve
 // to handle requests on incoming connections.
 //
-// If s.Addr is blank and LMTP is disabled, ":smtp" is used.
```
emersion: This change doesn't belong to this PR.

mschneider82: If we dont support TCP for lmtp, then the README.md example with telnet is wrong.

emersion: Why would it be wrong? The current README example creates a TCP server.

mschneider82: With my patch i have enabled TCP mode for LMTP, this was not allowed before and was forced to unixsocket. 
I have created an example for a LMTP TCP server with telnet as client in README.md, this example would be wrong if we dont accept TCP Mode for lmtp here.

mschneider82:
Lets discuss how we can go ahead here, I made much effort to get this ready and tested. 

Currently the LMTP support this in library is missing the "lmtp" feature to accept/not accept single rcpts, because of this feature lmtp was designed.

My patch is of couse breaking the existing API, but there is no way to not break it, because the current api is too simple for this kind of usage. 

Returning a slice of []error is breaking the RFC because it should be "near real-time". (also may breaks the api). A other solution sould be to call Data() for each rcpt but this is too strange.

I refactored the new interfaace to a general DataContext, this may can be used in future also for SMTP (e.g. context cancelation, getting all []recipients or something else).

emersion:
First of all, I appreciate your work on this PR, thanks for this.

I'm fine with breaking the API. However here are the things I dislike with this PR:

* Does too much. TCP for LMTP should be discussed separately. Context support should be added separately.
* `StartDelivery` is unnecessary
* I'd like to keep the README small, I don't want to clutter it with another server example.

mschneider82:
Ok, let's talk about 'StartDelivery is unnecessary': we could create the rcpt channels silent in the backend but then we have no possibility to add the ctx to it. The ctx is important to not block the channels and have a second way to get a status. 

I actually had other ways implemented and tested, also with performance tests

mschneider82:
I found another use case for the `DataContext`, i have implemented [XFORWARD](http://www.postfix.org/XFORWARD_README.html) support in my [feature branch](https://github.com/mschneider82/go-smtp/commits/feature-xforward)

Through the DataContext you can get the XForward informations from the mail.

foxcpp:
>I found another use case for the DataContext, i have implemented XFORWARD support in my feature branch

Side note: Since XFORWARD info doesn't change during transaction, it can be passed in ConnectionState structure to AnonymousLogin/Login.

Btw, what's the status of this PR?

I agree with emersion concerns about LMTP/TCP and README.
Can you revert both changes?

Also I believe that context support can be dropped altogether. I don't see any direct use for it and would like to keep library interfaces free from unnecessary details. Also, I think context-based timeouts and cancellation can be implemented in the backend without any help from the library.

mschneider82:
> The XFORWARD attributes are reset after the MAIL FROM transaction completes (after RSET or DOT), so there is no risk of information leakage.

mschneider82:
I suggest that i close this pull request, since my solution is not compatible with your vision of the library. I have to live with my fork. I'd like to keep the implementation of a server simple and not the library.

With your design every implementation of a lmtp server needs to handle with channels and status, timeouts (also the order of the status replies), this is too advanced in my opinion for an implementation.

Please use this as a base of better lmtp support. you can also check my master branch for other ideas (like the embbeded smtp.DefaultSession implementation)

Thanks for the library and your ideas.

foxcpp:
Closing due to previous comment by @mschneider82. I will submit "good" bits as my own PR.

PR #57, opened 2019-07-26
do not abort mailfrom if FROM is <> 

mschneider82:
The change was rebased with changes from #55

A FROM Adress <> is valid and used for NDR Reports / MailerDaemon mails.


emersion:
(Side note, next time you can just force-push to your branch instead of opening a new PR)

PR #58, opened 2019-08-07
Graceful shutdown by calling Close()

mschneider82:
fixed "use of closed network connection"

ListenAndServe() always returned an error when a Close() was called.
Since the returned error value is mostly calling a panic it prevents a clean shutdown.


foxcpp:
I wonder whether simply comparing error message to "use of closed connection" is better.

emersion:
>I wonder whether simply comparing error message to "use of closed connection" is better.

Hmm, please no.

PR #61, opened 2019-08-27
add come server documentation

kmpm:
Don't know if it's wanted but it would have saved me a some time


emersion:
This would be a better fit for the wiki. I just enabled it, can you post your docs there?

kmpm:
Sure I can. Wiki sounds like a perfect spot.

PR #66, opened 2019-09-26
server: Implement per-recipient status for LMTP

foxcpp:
Main differences from #56 are mentioned in the commit message:

>A separate interface is used so backend implementation will be aware
>of LMTP usage. This also avoids clutter in simple backend
>implementations that don't support LMTP.

Though I am not confident about this. Perhaps, having an argument for the Session.Data that is set to nil for non-LMTP would work better?

>RFC 2033 (Section 4.2) says that we should send multiple status lines if
>the same address was specified multiple times. We move responsibility of
>doing this to the backend implementation and don't do any deduplication
>themselves. This allows transparent forwarding for LMTP to be
>implemented correctly.

And this PR does not mess with LMTP/TCP or contexts, of course.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -466,6 +466,8 @@ func (c *Conn) handleData(arg string) {
 	// We have recipients, go to accept data
 	c.WriteResponse(354, EnhancedCode{2, 0, 0}, "Go ahead. End your data with <CR><LF>.<CR><LF>")
 
+	defer c.reset()
+
 	if c.server.LMTP {
 		c.handleDataLMTP()
 		return
@@ -476,7 +478,6 @@ func (c *Conn) handleData(arg string) {
 	io.Copy(ioutil.Discard, r) // Make sure all the data has been consumed
 	c.WriteResponse(code, enhancedCode, msg)
 
-	c.reset()
 }
 
 type statusCollector struct {
--- a/conn.go
+++ b/conn.go
@@ -481,15 +481,9 @@ func (c *Conn) handleData(arg string) {
 }
 
 type statusCollector struct {
-	mapLock sync.Mutex
-
 	// Contains map from recipient to list of channels that are used for that
 	// recipient.
-	//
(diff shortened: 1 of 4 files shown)


foxcpp:
So, @emersion, what's up this one?

conn.go
```diff
@@ -466,40 +466,122 @@ func (c *Conn) handleData(arg string) {
...
+type statusCollector struct {
+	mapLock sync.Mutex
+
+	// Contains map from recipient to list of channels that are used for that
+	// recipient.
+	//
+	// First SetStatus call uses first channel, second - second, etc. Channels
+	// that are already used are set to nil (otherwise we can accidentally
+	// reuse channel that was consumed by handleDataLMTP already).
+	statusMap map[string][]chan error
```
emersion: I'm not sure I understand why we need such a complicated data structure. Wouldn't a single channel be enough, with a `map[string]bool` on the receiver side to keep track of recipients we got status for?

foxcpp: The same recipient can be specified N times and we should send N status codes for it, possibly different ones. Though I don't remember why I used `[]chan error`, perhaps it should be a single channel.

emersion: Maybe then use a `map[string]int` to keep track of how many status replies we've sent for a recipient? (And still only one channel for everybody)

foxcpp: What's about order then? Who should be responsible for making sure responses are sent in the correct order (same as RCPT TO order)?

foxcpp: If we consider backends that do delivery for different recipients in parallel, it might be useful for go-smtp to keep track of the order instead (what this PR does, btw).

`map[string]chan error` alongside `[]chan error` looks good enough for me to solve this. With a single channel for everybody we need to do something when we receive statuses in a different order. Probably by putting statuses we can't send now into map[string][]error and using them later. What do you think?

foxcpp:
Aw, there is also a bug with copied lock in statusCollector.

PR #69, opened 2019-10-26
Better support for optional MAIL Arguments

### LINKED ISSUE #62: SMTP REQUIRETLS support
Basically, there is new argument to the MAIL FROM command (REQUIRETLS), it need to be parsed and passed to the backend. Also capability REQUIRETLS needs to be specified, but only when backend can actually make use of REQUIRETLS argument.

The message should be rejected immediately if REQUIRETLS argument is specified and the connection is not using TLS.

Related to foxcpp/maddy#123.
IETF I-D: https://tools.ietf.org/html/draft-ietf-uta-smtp-require-tls-09

### LINKED ISSUE #67: SMTPUTF8 capability
I believe most backend implementations could handle UTF-8 addresses and MIME headers just fine, so it is a good idea to tell clients that we are fine with UTF-8.

Also I think we should tell the backend that SMTPUTF8 support was requested (through MAIL argument). This way it can make decision on whether it can actually handle the message (say, if it implements transparent forwarding and upstream server doesn't support SMTPUTF8).

References:
- SMTP Extension for SMTPUTF8: https://tools.ietf.org/html/rfc6531
- Internationalized Email Headers: https://tools.ietf.org/html/rfc6532

foxcpp:
I decided to go with MailExt for client instead of adding an argument to Mail
because most users probably don't care about optional arguments.

Closes #62.
Closes #67.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -21,6 +21,13 @@ type Backend interface {
 	AnonymousLogin(state *ConnectionState) (Session, error)
 }
 
+// MailOptions contains custom arguments that were
+// passed as an argument to the MAIL command.
+type MailOptions struct {
+	// Size of the body. Can be 0 if not specified by client.
+	Size int
+}
+
 type Session interface {
 	// Discard currently processed message.
 	Reset()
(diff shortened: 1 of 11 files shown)


emersion:
>I decided to go with MailExt for client instead of adding an argument to Mail
because most users probably don't care about optional arguments.

Maybe we can take a pointer to options in `Mail` and accept nil?

foxcpp:
>Maybe we can take a pointer to options in Mail and accept nil?

No objection, will do.

PR #77, opened 2019-11-29
server: Rewrite MAIL FROM arguments handling 

foxcpp:
SMTPUTF8 and REQUIRETLS arguments are specified without the '=' sign.

Note to myself: If you think you can get away without writting tests for
something, you are terribly wrong and should suffer.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -302,26 +302,38 @@ func (c *Conn) handleMail(arg string) {
 			return
 		}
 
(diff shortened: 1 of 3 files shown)


emersion:
Err, tests are failing. BTW, would be nice to add a test for `SMTPUTF8`.

foxcpp:
I guess this is even better.

PR #80, opened 2019-12-07
client: Restrict the line length for responses

foxcpp:
Intended to protect MTA implementations from malicious recipient servers.

I am not sure about the length limit value though, for Server it is
configurable via the MaxLineLength field, for Client I do not see a context
where it can be set. Or perhaps it is not really needed?

### DIFF
--- a/client.go
+++ b/client.go
@@ -65,7 +65,21 @@ func DialTLS(addr string, tlsConfig *tls.Config) (*Client, error) {
 // NewClient returns a new Client using an existing connection and host as a
 // server name to be used when authenticating.
 func NewClient(conn net.Conn, host string) (*Client, error) {
-	text := textproto.NewConn(conn)
+	rwc := struct {
+		io.Reader
+		io.Writer
+		io.Closer
+	}{
+		Reader: lineLimitReader{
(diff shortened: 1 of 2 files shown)


emersion:
The RFC says the maximum length for replies is 512 bytes (https://tools.ietf.org/html/rfc5321#section-4.5.3.1.5). I'm fine with hardcoding this, at least for now.

emersion:
Note, I just realized `textproto.Conn` has a `Reader` field, which has a `bufio.Reader` field, which allows us to set the maximum buffer size.

foxcpp:
>bufio.Reader field, which allows us to set the maximum buffer size.

Does it? The only 'size' I see is in NewReaderSize which sets *minimal* buffer size.

Additionally, textproto.Conn.ReadLine* joins results of multiple R.Read calls so it is not limited by the bufio.Reader buffer (see readLineSlice implementation in net/textproto/reader.go)

emersion:
>Does it? The only 'size' I see is in NewReaderSize which sets minimal buffer size.

Oh, I'm mistaken, I was thinking about `bufio.Scanner` instead.

PR #92, opened 2020-04-26
Add Client.LMTPData method with ability to retrieve per-recipient status

### LINKED ISSUE #91: Allow LMTP client to retrieve per-recipient DATA status
I think providing an interface similar to [LMTPSession](https://pkg.go.dev/github.com/emersion/go-smtp?tab=doc#LMTPSession) on the client side is a good solution.

Follow-up for #23.
Ref. foxcpp/maddy#205

foxcpp:
Closes #91.

Ref.: foxcpp/maddy#205

### DIFF
--- a/client.go
+++ b/client.go
@@ -365,6 +365,7 @@ func (c *Client) Rcpt(to string) error {
 type dataCloser struct {
 	c *Client
 	io.WriteCloser
+	statusCb func(status *SMTPError)
 }
 
 func (d *dataCloser) Close() error {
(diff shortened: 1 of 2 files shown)


emersion:
Code LGTM, but this interface sounds a little bit dangerous. Could we:

* Check if the client is LMTP, error out if not?
* Record the list of recipients an add it as a parameter to the callback?

foxcpp:
>     * Check if the client is LMTP, error out if not?
> 
>     * Record the list of recipients an add it as a parameter to the callback?

Done.

PR #96, opened 2020-05-09
Properly handle AUTH= parameter for MAIL command

### LINKED ISSUE #95: MAIL in Strict mode doesn't support optional extensions
It looks like the MAIL parsing in strict mode expects the only argument to be the FROM argument, but the SMTP spec allows for optional `mail-parameters` that are associated with service extensions. See https://tools.ietf.org/html/rfc5321#section-3.3.

It might be safe to ignore these parameters, but right now the server returns a 501 because it expects the argument to end with a `>`.

Example command from the client:

```
MAIL FROM:<example@example.com> AUTH=<username>
```

Response from go-smtp:

```
501 Was expecting MAIL arg syntax of FROM:<address>
```

It seems like KMail includes these additional parameters (see https://www.reddit.com/r/ProtonMail/comments/evadaz/need_help_setting_up_outgoing_with_kmail/), as well as [emailrelay](http://emailrelay.sourceforge.net/index.html).

foxcpp:
This PR adds both client-side and server-side support.
Closes #95.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -3,11 +3,13 @@ package smtp
 import (
 	"crypto/tls"
 	"encoding/base64"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"net"
 	"net/textproto"
+	"regexp"
 	"runtime/debug"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


conn.go
```diff
@@ -358,6 +376,57 @@ func (c *Conn) handleMail(arg string) {
...
+func encodeXtext(raw string) string {
+	var out strings.Builder
+	out.Grow(len(raw))
+
+	for _, ch := range raw {
+		if ch == '+' || ch == '=' {
+			out.WriteRune('+')
+			out.WriteString(strings.ToUpper(strconv.FormatInt(int64(ch), 16)))
+		}
+		if ch > '~' && ch < '~' { // printable non-space US-ASCII
```
foxcpp, line 420: Hell.

PR #100, opened 2020-05-18
Added SetLogHandler method on client.

### LINKED ISSUE #48: Allow to log raw I/O in Client
Like Server.Debug but for Client.

discobean:
I believe it fixes #48

I just wanted a way to specify more detailed logging.

Use it by 
```
c.SetLogHandler(func(message string) {
  log.Print(message)
})
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,8 @@
 module github.com/emersion/go-smtp
 
-require github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21
+require (
+	github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21
+	github.com/sirupsen/logrus v1.6.0 // indirect
+)
 
 go 1.13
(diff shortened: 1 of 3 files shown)


foxcpp:
For referenced issue I had the idea of exactly replicating `Debug io.Writer` of server code by wrapping net.Conn with io.TeeWriter. Do you think callback is a better approach?

discobean:
Most definitely replicating io.Writer is better, but this was just a quick way to get it over the line for me.

go.mod
```diff
@@ -1,5 +1,8 @@
 module github.com/emersion/go-smtp
 
-require github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21
+require (
+	github.com/emersion/go-sasl v0.0.0-20200509203442-7bfe0ed36a21
+	github.com/sirupsen/logrus v1.6.0 // indirect
```
emersion, line 5: We don't need this dep

emersion:
I'd prefer an `io.Writer` as well.

Issue #81, opened 2019-12-07
RFC 3030 (CHUNKING, BINARYMIME) support


foxcpp:
## CHUNKING (BDAT) support

Client support: Prefer over DATA if available (avoids the overhead of dot-encoding), require if BODY=BINARYMIME is used (as required by spec).
Server support: Transparent support, io.Reader passed to Session.Data reads multiple chunks. Error returned by Data early consumes the current chunk and returns an error without consuming the whole message.

## BINARYMIME support

Support BODY=BINARYMIME argument for MAIL command, no other actions required.

Specification: https://tools.ietf.org/html/rfc3030

emersion:
Server part has been merged in https://github.com/emersion/go-smtp/pull/106. Keeping this issue opened for client support.

norguhtar:
I'm test BINARYMIME and get error. What i do. Send mail with attachment (10mb size) like this

```
BDAT 1048576
...
some binary data
....
554 5.0.0 Error: transaction failed, blame it on the weather: smtp: too longer line in input stream
```

I checked source and found this error generated LimitReader when readLine is got limit. But when we used BDAT and BINARYMIME content not split by lines

From RFC3030
````
 When the receiver-SMTP accepts a
   MAIL command with the BINARYMIME body-value, it agrees to preserve
   all bits in each octet passed using the BDAT command.  Once a
   receiver-SMTP supporting the BINARYMIME service extension accepts a
   message containing binary material, the receiver-SMTP MUST deliver or
   relay the message in such a way as to preserve all bits in each
   octet.
````

I can set bigger MaxLineLength but think this not really right solution

foxcpp:
Oh well, this is a problem. Perhaps we could disable lineLimitReader while reading a BDAT chunk? Can you make a PR?

emersion:
Looks like a good idea to me too.

norguhtar:
> Oh well, this is a problem. Perhaps we could disable lineLimitReader while reading a BDAT chunk? Can you make a PR?

Yes. I investigate problem and send PR

emersion:
What would be a good chunk size for BDAT? Should we only use BDAT when PIPELINING is also available (to avoid having to wait for the server ACK'ing each chunk)?

norguhtar:
I'm send really big data. For work purpose i'm need send 8Gb data file. And use really big chunk size. I think usefull size 1Megabyte.

PR #104, opened 2020-06-11
Server-side CHUNKING and BINARYMIME support

foxcpp:
See #81. This issue is not closed since it also references client support.

### DIFF
--- a/chunk_reader.go
+++ b/chunk_reader.go
@@ -19,7 +19,7 @@ var ErrDataReset = errors.New("smtp: message transmission aborted")
 // end of another chunk being reached.
 type chunkReader struct {
 	remainingBytes int
-	conn           io.Reader
+	r              io.Reader
 	chunks         chan int
 	// Sent to by abort() to unlock running Read.
 	rset         chan struct{}
@@ -58,7 +58,7 @@ func (cr *chunkReader) waitNextChunk() error {
 			// Okay, that's the end.
 			return io.EOF
 		}
-		cr.currentChunk = &io.LimitedReader{R: cr.conn, N: int64(r)}
+		cr.currentChunk = &io.LimitedReader{R: cr.r, N: int64(r)}
 		return nil
 	}
 }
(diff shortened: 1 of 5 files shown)


chunk_reader.go
```diff
@@ -0,0 +1,127 @@
...
+// chunkReader implements io.Reader by consuming a sequence of size-limited
+// "chunks" from underlying Reader allowing them to be interleaved with other
+// protocol data.
+//
+// It is caller responsibility to not use Reader while chunk is being processed.
+// This can be enforced by blocking on chunkEnd channel that is used to signal the
+// end of another chunk being reached.
+type chunkReader struct {
+	remainingBytes int
+	conn           io.Reader
```
corny, line 22: What do you thing about renaming `conn` to `r`, `reader`, or `input`?

foxcpp: Done

chunk_reader.go
```diff
@@ -0,0 +1,128 @@
...
+	return n - offset, err
+
+}
+
+func newChunkReader(conn io.Reader, maxBytes int) *chunkReader {
+	return &chunkReader{
+		remainingBytes: maxBytes,
+		r:              conn,
+		chunks:         make(chan int, 1),
+		// buffer to make sure abort() will not block if Read is not running.
```
corny, line 124: abort() will never block because it just closes the channels.
Channels that are never written to don't need a buffer.

foxcpp: Removed comment in e797215553b133513fa3c7bd3a43bd9f64699621.

foxcpp:
> Instead of this `chunkReader` approach, can't we read data directly from `handleBdat` and forward it to the backend via an `io.Pipe`?

I wanted to preserve ability for backend to return errors without consuming the whole message but I will look if this is easily possible with io.Pipe.

emersion:
>I wanted to preserve ability for backend to return errors without consuming the whole message but I will look if this is easily possible with io.Pipe.

`io.PipeReader.CloseWithError` could help.

PR #106, opened 2020-06-23
Server-side CHUNKING and BINARYMIME support

foxcpp:
See #81. This issue is not closed since it also references client support.
Replaces #104 (but hey, that initial solution saves one message copy :) )

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -21,9 +21,20 @@ type Backend interface {
 	AnonymousLogin(state *ConnectionState) (Session, error)
 }
 
+type BodyType string
+
+const (
+	Body7Bit       BodyType = "7BIT"
+	Body8BitMIME   BodyType = "8BITMIME"
+	BodyBinaryMIME BodyType = "BINARYMIME"
+)
+
 // MailOptions contains custom arguments that were
 // passed as an argument to the MAIL command.
 type MailOptions struct {
+	// Value of BODY= argument, 7BIT, 8BITMIME or BINARYMIME.
+	Body BodyType
+
 	// Size of the body. Can be 0 if not specified by client.
 	Size int
 
--- a/conn.go
+++ b/conn.go
@@ -260,6 +260,9 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 		if _, isTLS := c.TLSConnectionState(); isTLS && c.server.EnableREQUIRETLS {
 			caps = append(caps, "REQUIRETLS")
 		}
+		if c.server.EnableBINARYMIME {
+			caps = append(caps, "BINARYMIME")
+		}
 		if c.server.MaxMessageBytes > 0 {
 			caps = append(caps, fmt.Sprintf("SIZE %v", c.server.MaxMessageBytes))
 		}
@@ -350,11 +353,17 @@ func (c *Conn) handleMail(arg string) {
 				opts.RequireTLS = true
 			case "BODY":
 				switch value {
+				case "BINARYMIME":
+					if !c.server.EnableBINARYMIME {
+						c.WriteResponse(504, EnhancedCode{5, 5, 4}, "BINARYMIME is not implemented")
+						return
+					}
 				case "7BIT", "8BITMIME":
 				default:
 					c.WriteResponse(500, EnhancedCode{5, 5, 4}, "Unknown BODY value")
 					return
 				}
+				opts.Body = BodyType(value)
 			case "AUTH":
 				value, err := decodeXtext(value)
 				if err != nil {
--- a/conn.go
+++ b/conn.go
@@ -33,6 +33,11 @@ type Conn struct {
 	session   Session
 	locker    sync.Mutex
 
+	bdatPipe      *io.PipeWriter
+	bdatStatus    *statusCollector // used for BDAT on LMTP
+	dataResult    chan error
+	bytesReceived int // counts total size of chunks when BDAT is used
+
 	fromReceived bool
 	recipients   []string
 }
@@ -133,6 +138,8 @@ func (c *Conn) handle(cmd string, arg string) {
 	case "RSET": // Reset session
 		c.reset()
 		c.WriteResponse(250, EnhancedCode{2, 0, 0}, "Session reset")
+	case "BDAT":
+		c.handleBdat(arg)
 	case "DATA":
 		c.handleData(arg)
 	case "QUIT":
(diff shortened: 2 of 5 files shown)


conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+		// src stopped early; must have been EOF.
+		err = io.EOF
+	}
+
+	return consumed, err
+}
+
+func (c *Conn) handleBdat(arg string) {
+	args := strings.Split(arg, " ")
+	if len(args) == 0 {
```
emersion, line 702: `strings.Split` never returns an empty slice

conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+		return
+	}
+
+	if !c.fromReceived || len(c.recipients) == 0 {
+		c.WriteResponse(502, EnhancedCode{5, 5, 1}, "Missing RCPT TO command.")
+		return
+	}
+
+	last := false
+	if len(args) == 2 {
```
emersion, line 713: Do we want to be strict or not?

- If we want to be strict, use `strings.SplitN` to avoid having `len(args) > 2`
- If we want to be lax, change this check to `len(args) >= 2` to fail on `BDAT 42 INVALID INVALID`

conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+
+	if c.bdatStatus == nil && c.server.LMTP {
+		c.bdatStatus = c.createStatusCollector()
+	}
+
+	if c.bdatPipe == nil {
+		var r *io.PipeReader
+		r, c.bdatPipe = io.Pipe()
+
+		c.dataResult = make(chan error)
```
emersion, line 736: I'm not sure I understand the point of this goroutine. What is it used for?

I'm a little confused because it's read from and written to in many different goroutines.

conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+			}
+
+			select {
+			case <-c.dataResult:
+				// handleBdat goroutine wants us to sent an error here.
+				c.dataResult <- err
+			default:
+			}
+
+			if err != nil {
```
emersion, line 775: Nit: `r.CloseWithError(nil)` is the same as `r.Close()`

conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+			}
+		}()
+	}
+
+	// Using copyN that returns amount of bytes read from c.text.R instead of
+	// amount of bytes written is important here because read may succeed and
+	// write not in case of early error returned by Session.Data. In this case
+	// io.CopyN call below needs to calculate correct amount of bytes to
+	// discard.
+	consumed, err := copyN(c.bdatPipe, c.text.R, int64(size))
```
emersion, line 788: Instead of re-implementing `copyN`, can't we use `io.LimitedReader` to create a reader that'll EOF after `size` bytes? Then we can safely copy it, and discard till EOF on error.

conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+		if err == errPanic {
+			c.Close()
+		}
+
+		return
+	}
+
+	c.bytesReceived += int(consumed)
+	if c.server.MaxMessageBytes != 0 && c.bytesReceived > c.server.MaxMessageBytes {
+		c.WriteResponse(552, EnhancedCode{5, 3, 4}, "Max message size exceeded")
```
emersion, line 805: Should we close the `bdatPipe` at this point?

Maybe we can do this before actually copying the data?

conn.go
```diff
@@ -604,7 +647,239 @@ func (c *Conn) handleData(arg string) {
...
+
+	if last {
+		c.bdatPipe.Close()
+
+		// At this point Session.Data might still be running and processing the
+		// message we need to a) wait for it before killing the session b)
+		// obtain the error if any.
+		//
+		// io.Pipe CloseWithError cannot be used at this point since we are not
+		// writing anymore. Therefore we signal that we want an error sent to this
```
emersion, line 817: Hmm. So this is the explanation for https://github.com/emersion/go-smtp/pull/106/files#r447675105, right?

Why do we need to signal to the bdat goroutine that we want an error? Can't the goroutine always send one? (We can change the channel to `make(chan error, 1)` to avoid leaking the bdat goroutine if we don't read the error here.)

conn.go
foxcpp: Right, that check should be moved before io.Copy.

conn.go
foxcpp: Replacing that call with `strings.Fields` (and adding explicit `len(args) > 2` check) seems to be clean solution to me.

conn.go
foxcpp: That works, will do.

conn.go
foxcpp: You are right about this one too, dunno what I was thinking when writing that.

backend.go
```diff
@@ -24,6 +24,9 @@ type Backend interface {
 // MailOptions contains custom arguments that were
 // passed as an argument to the MAIL command.
 type MailOptions struct {
+	// Value of BODY= argument, 7BIT, 8BITMIME or BINARYMIME.
+	Body string
```
emersion, line 28: Should we provide a type and constants for valid values?

conn.go
```diff
@@ -604,7 +647,182 @@ func (c *Conn) handleData(arg string) {
...
+		c.dataResult = make(chan error, 1)
+
+		go func() {
+			defer func() {
+				if err := recover(); err != nil {
+					c.handlePanic(err, c.bdatStatus)
+
+					select {
+					case <-c.dataResult:
+						c.dataResult <- errPanic
```
emersion, line 707: We can unconditionally send the error to the `dataResult` channel here I think.

conn.go
```diff
@@ -604,7 +647,182 @@ func (c *Conn) handleData(arg string) {
...
+	}
+
+	chunk := io.LimitedReader{R: c.text.R, N: int64(size)}
+	_, err = io.Copy(c.bdatPipe, &chunk)
+	if err != nil {
+		// Backend might return an error without consuming the whole chunk.
+		io.Copy(ioutil.Discard, &chunk)
+
+		code, enhancedCode, msg := toSMTPStatus(err)
+		c.WriteResponse(code, enhancedCode, msg)
```
emersion, line 742: Style nit: I think this works:

```go
c.WriteResponse(toSMTPStatus(err))
```

conn.go
```diff
@@ -604,7 +647,182 @@ func (c *Conn) handleData(arg string) {
...
+		}
 
+		c.reset()
+		return
+	}
+
+	c.bytesReceived += int(size)
+
+	if last {
+		c.bdatPipe.Close()
```
emersion, line 755: Note to self: pipes never return an error when closed.

emersion:
Also please squash the commits into logical commits.

conn.go
foxcpp: You are right, I forgot to update this path when copy-pasting code from previous PR.

foxcpp:
I just realized I have local changes I forgot to commit that resolve some of issues you mentioned. :)

PR #111, opened 2020-07-13
Replace net/textproto DotReader with custom implementation to avoid CRLF->LF rewrite

### LINKED ISSUE #108: server: Switch away from using textproto.DotReader
DotReader does CRLF->LR conversion that is generally unwanted for most email processing needs.

foxcpp:
I copy-pasted net/textproto code and I am not completely sure how it works.
If you want me to improve it - I can try though :)

Closes #108.

### DIFF
--- a/data.go
+++ b/data.go
@@ -1,6 +1,7 @@
 package smtp
 
 import (
+	"bufio"
 	"io"
 )
 
@@ -42,7 +43,8 @@ var ErrDataTooLarge = &SMTPError{
 }
 
 type dataReader struct {
-	r io.Reader
+	r     *bufio.Reader
+	state int
 
 	limited bool
 	n       int64 // Maximum bytes remaining
@@ -50,7 +52,7 @@ type dataReader struct {
 
 func newDataReader(c *Conn) *dataReader {
 	dr := &dataReader{
-		r: c.text.DotReader(),
+		r: c.text.R,
 	}
 
 	if c.server.MaxMessageBytes > 0 {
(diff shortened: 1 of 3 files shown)


data.go
```diff
@@ -71,7 +73,86 @@ func (r *dataReader) Read(b []byte) (n int, err error) {
...
+				break
+			}
+			// Not part of \r\n. Emit saved \r
+			r.r.UnreadByte()
+			c = '\r'
+			r.state = stateData
+
+		case stateData:
+			if c == '\r' {
+				r.state = stateCR
```
emersion, line 144: Note to self: the difference with the upstream implementation is that we don't `continue` here.

data.go
```diff
@@ -71,7 +73,86 @@ func (r *dataReader) Read(b []byte) (n int, err error) {
...
+			r.state = stateData
+
+		case stateDotCR:
+			if c == '\n' {
+				r.state = stateEOF
+				continue
+			}
+			// Not part of .\r\n.
+			// Consume leading dot and emit saved \r.
+			r.r.UnreadByte()
```
emersion, line 128: We don't "save" the `\r` anymore, since we don't skip appending to `b` here: https://github.com/emersion/go-smtp/pull/111/files#diff-73850c2e1fb203a205030b007f382b14R144

So, I believe this trips on `.\rasdf`, and emits a duplicate `\r`. I think we can fix it by just updating state to `stateData` without calling `UnreadByte` and without changing the value of `c`.

data.go
```diff
@@ -71,7 +73,86 @@ func (r *dataReader) Read(b []byte) (n int, err error) {
...
+			c = '\r'
+			r.state = stateData
+
+		case stateCR:
+			if c == '\n' {
+				r.state = stateBeginLine
+				break
+			}
+			// Not part of \r\n. Emit saved \r
+			r.r.UnreadByte()
```
emersion, line 138: Ditto

data.go
corny: Please paste this comment into the code. Could be also important for others.

emersion: Well, there are now more differences. But OTOH the code is now easier to read :)

PR #112, opened 2020-07-14
server: Move to using Backend.Features() instead of Enable* flags

### LINKED ISSUE #107: Replace Server.Enable* with backend Features() method
```go
type Feature int
const (
   FeatureSMTPUTF8 = 1<<iota
   FeatureREQUIRETLS
   FeatureBINARYMIME
)

// ...
interface Backend {
  // ...
  Features() Feature
}
```

foxcpp:
Closes #107.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -21,6 +21,18 @@ type Backend interface {
 	AnonymousLogin(state *ConnectionState) (Session, error)
 }
 
+type Feature int32
+
+const (
+	FeatureNoSMTPUTF8 Feature = 2 << iota
+	FeatureNoBINARYMIME
+	FeatureNoREQUIRETLS
+)
+
+type FeatureBackend interface {
+	Features() Feature
+}
+
 type BodyType string
 
 const (
--- a/conn.go
+++ b/conn.go
@@ -255,13 +255,14 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 
 			caps = append(caps, authCap)
 		}
-		if c.server.EnableSMTPUTF8 {
+		feat := c.server.backendFeatures()
+		if feat&FeatureNoSMTPUTF8 == 0 {
 			caps = append(caps, "SMTPUTF8")
 		}
-		if _, isTLS := c.TLSConnectionState(); isTLS && c.server.EnableREQUIRETLS {
+		if _, isTLS := c.TLSConnectionState(); isTLS && feat&FeatureNoREQUIRETLS == 0 {
 			caps = append(caps, "REQUIRETLS")
 		}
-		if c.server.EnableBINARYMIME {
+		if feat&FeatureNoBINARYMIME == 0 {
 			caps = append(caps, "BINARYMIME")
 		}
 		if c.server.MaxMessageBytes > 0 {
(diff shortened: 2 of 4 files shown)


corny:
I would like to see a `func (f Feature) Contains(feat int) bool { ... }` method to clean up the code.

backend.go
```diff
@@ -21,6 +21,18 @@ type Backend interface {
 	AnonymousLogin(state *ConnectionState) (Session, error)
 }
 
+type Feature int32
+
+const (
+	FeatureNoSMTPUTF8 Feature = 2 << iota
```
emersion, line 27: I'm not a big fan of these `NoXXX` "features". Why the inversion?

emersion:
I wonder how we should approach backwards compatibility. If we add a new feature, we won't be able to enable it by default without breaking the backend?

foxcpp:
Addressed feedback, added Feature.Contains() and inverted flags (making extensions opt-in rather than opt-out).

backend.go
```diff
@@ -21,6 +21,22 @@ type Backend interface {
 	AnonymousLogin(state *ConnectionState) (Session, error)
 }
 
+type Feature int32
```
emersion, line 24: Can we make this an `uint32`?

backend.go
```diff
@@ -21,6 +21,22 @@ type Backend interface {
...
+
+const (
+	FeatureSMTPUTF8 Feature = 2 << iota
```
emersion, line 27: Can we keep the references to the RFCs that were in the `Server` fields?

backend.go
```diff
@@ -21,6 +21,22 @@ type Backend interface {
...
+	FeatureBINARYMIME
+	FeatureREQUIRETLS
+)
+
+func (f Feature) Contains(feat Feature) bool {
+	return f&feat != 0
+}
+
+type FeatureBackend interface {
```
emersion, line 36: Maybe embed the `Backend` interface here, to make it clear it's a backend extension

emersion (changes requested):
This now ensures adding new features to go-smtp won't break any backend, but it's not very ergonomic for users...

backend.go
foxcpp: Done.

backend.go
foxcpp: Done.

backend.go
foxcpp: Done.

backend.go
```diff
@@ -21,6 +21,26 @@ type Backend interface {
 	AnonymousLogin(state *ConnectionState) (Session, error)
 }
 
+type Feature uint32
+
+const (
+	// SMTPUTF8 (RFC 6531) extension.
+	FeatureSMTPUTF8 Feature = 2 << iota
```
emersion, line 28: Why is this 2 instead of 1?

backend.go
```diff
@@ -21,6 +21,26 @@ type Backend interface {
...
+	// BINARYMIME (RFC 3030) extension. CHUNKING extension is always supported.
+	FeatureBINARYMIME
```
emersion, line 30: The rest of the code seems to use the Go capitalization: "BinaryMIME"

(Same for "RequireTLS" I suppose)

backend.go
foxcpp: Ouch, most likely a typo

PR #114, opened 2020-07-15
server: XCLIENT implementation

foxcpp:
Some interface details probably need discussion.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -73,6 +73,24 @@ type Session interface {
 	Data(r io.Reader) error
 }
 
+type ProxyBackend interface {
+	// AllowProxy method is called when client uses XCLIENT command without
+	// initialized session.
+	//
+	// Backends should implement this method AND returned Session objects
+	// should implement ProxySession for XCLIENT to work correctly.
+	AllowProxy(actual, asserted ConnectionState) bool
+}
+
+type ProxySession interface {
+	// AllowProxy method is called when client uses XCLIENT command.
+	//
+	// This is similar to ProxyBackend.AllowProxy but called instead of it if
+	// Session is already created for the user. Session.Logout will be
+	// called if this function returns true.
+	AllowProxy(asserted ConnectionState) bool
+}
+
 type LMTPSession interface {
 	// LMTPData is the LMTP-specific version of Data method.
 	// It can be optionally implemented by the backend to provide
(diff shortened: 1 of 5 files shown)


emersion:
Hmm, this backend API is a little bit unfortunate. Would there be a way to only have a new `Session` method?

foxcpp:
> Hmm, this backend API is a little bit unfortunate. Would there be a way to only have a new `Session` method?

It is due to the fact XCLIENT is valid both before and after AUTH and may work differently depending on authentication status. I initially thought about avoiding two methods by creating a session using anonymous callback but it may be problematic since backend may permit XCLIENT but not unauthenticated transactions and permit them after XCLIENT.  ( On a second though that setup does not really make a lot of sense, not sure... )

If I had the liberty of doing a major interface change I would ditch authentication-centered Session creation and add Auth() method to Session, create one session per connection and be done with it (we already have one session per connection but the way it is initially created is rather convoluted).

emersion:
>If I had the liberty of doing a major interface change I would ditch authentication-centered Session creation and add Auth() method to Session, create one session per connection and be done with it (we already have one session per connection but the way it is initially created is rather convoluted).

Agreed. Do you think it's worth it to break the API this way?

foxcpp:
I believe HAProxy's PROXY protocol is more widely supported therefore it is a good idea to implement it first.

foxcpp:
It is also much more simple than XCLIENT which is SMTP-focused. It might be a good idea to support it too if we take into account SMTP-aware proxies.

kayrus:
HAProxy's PROXY protocol is easy to implement using the https://github.com/pires/go-proxyproto package:
```
                        // Wrap listener in a proxyproto listener
                        proxyListener := &proxyproto.Listener{Listener: smtpListener}
                        err := smtpServer.Serve(proxyListener)
```

foxcpp:
Closing as I do not intend to continue work on XCLIENT in foreseeable future.

foxcpp:
After a more careful research I made the decision it is useful to support both PROXY and XCLIENT since XCLIENT is more flexible and is compatible with connection pooling while PROXY is not.

xdkaka:
> HAProxy's PROXY protocol is easy to implement using the https://github.com/pires/go-proxyproto package:
> 
> ```
>                         // Wrap listener in a proxyproto listener
>                         proxyListener := &proxyproto.Listener{Listener: smtpListener}
>                         err := smtpServer.Serve(proxyListener)
> ```


In this way, is the proxy server able to forge the user's real IP?

Issue #71, opened 2019-11-14
Denial of service possible (DoS)


BenoitGEORGELIN:
Hello, 
I'm doing some testing on this package, thanks for the great work. :+1: 

I wanted to share with you two possible denial of service. 
Right now, anybody can open a connection and keep it for ever doing : 

**- Bad command ( commands that are not handle by the parse() function )**
c.nbrErrors++ will be incremented but will never trigger a connection close() 

**- Speak Up (When command is empty)** 
A connection can stay open , sending empty commands for ever 

In both cases, the attacker can open as many connection, send either "bad commands" not parsed. So , for example, 1 char, 2char, 3 char commands or just empty commands and will never reach any disconnect. 

As a quick fix, I would suggest this to evaluate the c.nbrErrors counter and disconnect if > 3
In both scenarios 

**Example:** 
conn.go
Line: 101
```
	if cmd == "" {
		c.nbrErrors++
		if c.nbrErrors > 3 {
			c.WriteResponse(500, EnhancedCode{5, 5, 2}, "Too many errors")
			c.Close()
		}	
		c.WriteResponse(500, EnhancedCode{5, 5, 2}, "Speak up")
		return
	}

```

file: server.go
Line: 137

```
	if err != nil {
		c.nbrErrors++
		if c.nbrErrors > 3 {
			c.WriteResponse(500, EnhancedCode{5, 5, 2}, "Too many errors")
			c.Close()
		}					
	
		c.WriteResponse(501, EnhancedCode{5, 5, 2}, "Bad command")
		continue
	}	

```

Hope this help.

foxcpp:
nbrErrors is checked here: https://github.com/emersion/go-smtp/blob/4bc96f6ca8dfb1eedae59385dec914dde09eaad3/conn.go#L81
But yeah, I think empty lines should be handled as invalid commands too. And nbrErrors should be checked in server.go:137. Perhaps, make a PR?

Also, if there's intention to DoS the server, this can't prevent the attacker from using valid (and potentially much more expensive) commands.

BenoitGEORGELIN:
In fact this check will only happen on commands with char  > 4 
I'll send you a PR

emersion:
Fixed in #116

PR #72, opened 2019-11-16
Add c.nbrErrors checks

BenoitGEORGELIN:
FIX: issue 71 - Denial of service possible (DoS) 

```
go-smtp$ go test
PASS
ok  	github.com/emersion/go-smtp	0.014s

```

PR #116, opened 2020-08-11
Improved error handling

nsmith5:
This pull requests makes error handling a little more consistent and fixes issue #71. Pull request #72 has been open to address this issue for a while but without much sign of life from the author. This change
- Factors the error handling into a function as suggested
- Parametrizes the error threshold and defaults it to 3 (could be configurable in the future? At least it lives in one place now)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -17,6 +17,9 @@ import (
 	"time"
 )
 
+// Number of errors we'll tolerate per connection before closing. Defaults to 3.
+const errThreshold = 3
+
 type ConnectionState struct {
 	Hostname   string
 	LocalAddr  net.Addr
@@ -32,8 +35,6 @@ type Conn struct {
 
 	// Number of errors witnessed on this connection
 	errCount int
-	// Number of errors we'll tolerate per connection before closing. Defaults to 3.
-	errThreshold int
 
 	session    Session
 	locker     sync.Mutex
@@ -50,9 +51,8 @@ type Conn struct {
 
 func newConn(c net.Conn, s *Server) *Conn {
 	sc := &Conn{
-		server:       s,
-		conn:         c,
-		errThreshold: 3,
+		server: s,
+		conn:   c,
 	}
 
 	sc.init()
@@ -225,7 +225,7 @@ func (c *Conn) protocolError(code int, ec EnhancedCode, msg string) {
 	c.WriteResponse(code, ec, msg)
 
 	c.errCount++
-	if c.errCount > c.errThreshold {
+	if c.errCount > errThreshold {
 		// 421 -> Connection closing (RFC5321 Section-4.2.3)
 		// 5.7.0 -> Persistent security related issue
 		c.WriteResponse(421, EnhancedCode{5, 7, 0}, "Too many errors. Quiting now")
--- a/conn.go
+++ b/conn.go
@@ -25,11 +25,16 @@ type ConnectionState struct {
 }
 
 type Conn struct {
-	conn       net.Conn
-	text       *textproto.Conn
-	server     *Server
-	helo       string
-	nbrErrors  int
+	conn   net.Conn
+	text   *textproto.Conn
+	server *Server
+	helo   string
+
+	// Number of errors witnessed on this connection
+	errCount int
+	// Number of errors we'll tolerate per connection before closing. Defaults to 3.
+	errThreshold int
+
 	session    Session
 	locker     sync.Mutex
 	binarymime bool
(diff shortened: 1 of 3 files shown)


conn.go
```diff
@@ -222,6 +219,18 @@ func (c *Conn) authAllowed() bool {
...
 }
 
+// handleError writes errors responses and closes the connection once too many
+// have occurred.
+func (c *Conn) handleError(code int, ec EnhancedCode, msg string) {
+	c.WriteResponse(code, ec, msg)
+
+	c.errCount++
+	if c.errCount > c.errThreshold {
+		c.WriteResponse(500, EnhancedCode{5, 5, 2}, "Too many errors. Quiting now")
```
foxcpp, line 229: RFC 5321 defines a special status code to use when connection is closed - 421. Would using it here be better?

conn.go
```diff
@@ -222,6 +219,18 @@ func (c *Conn) authAllowed() bool {
 	return !c.server.AuthDisabled && (isTLS || c.server.AllowInsecureAuth)
 }
 
+// handleError writes errors responses and closes the connection once too many
+// have occurred.
+func (c *Conn) handleError(code int, ec EnhancedCode, msg string) {
```
foxcpp, line 224: This name seems to imply it should be called for all errors. Perhaps it could be more specific? e.g. `protocolError`

conn.go
nsmith5: Would something like `c.WriteResponse(421, EnhancedCode{5, 7, 0}, "Too many errors. Quiting now")` work? Sort of stabbing about for that enhanced code. Persistent security related issue seems the most closely related to me.

conn.go
nsmith5: Sure, `protocolError` works for me.

conn.go
```diff
@@ -25,11 +25,16 @@ type ConnectionState struct {
...
-	nbrErrors  int
+	conn   net.Conn
+	text   *textproto.Conn
+	server *Server
+	helo   string
+
+	// Number of errors witnessed on this connection
+	errCount int
+	// Number of errors we'll tolerate per connection before closing. Defaults to 3.
+	errThreshold int
```
emersion, line 36: This can be a global constant

conn.go
foxcpp: First number in enhanced code should match the basic code, therefore it should be at least 4.7.0.
The list of registered codes can be found here: https://www.iana.org/assignments/smtp-enhanced-status-codes
I believe 4.5.1 would be the most meaningful.

foxcpp: First number in enhanced code should match the basic code, therefore it should be at least 4.7.0.
The list of registered codes can be found here: https://www.iana.org/assignments/smtp-enhanced-status-codes
I believe 4.5.1 would be the most meaningful.

conn.go
nsmith5: For sure. My thinking was that it could be configurable if that is wanted in the future, but I can move it out into a global for now.

conn.go
nsmith5: [RFC 3463 Section 3.6](https://tools.ietf.org/html/rfc3463#section-3.6) Specifies that `x.5.1` is useful only as a permenant error (so 5.5.1).

foxcpp: Okay, I think 421 is to be used when server closes the connection for some internal reason (e.g. shutdown). 500 5.5.1 seems fine to indicate the case of possibly malfunctioning client. I had some concerns about how permanently failing delivery (5xx) would affect such cases but I think it is fine.

conn.go
```diff
@@ -222,6 +219,20 @@ func (c *Conn) authAllowed() bool {
...
+// protocolError writes errors responses and closes the connection once too many
+// have occurred.
+func (c *Conn) protocolError(code int, ec EnhancedCode, msg string) {
+	c.WriteResponse(code, ec, msg)
+
+	c.errCount++
+	if c.errCount > errThreshold {
+		// 421 -> Connection closing (RFC5321 Section-4.2.3)
+		// 5.7.0 -> Persistent security related issue
+		c.WriteResponse(421, EnhancedCode{5, 7, 0}, "Too many errors. Quiting now")
```
foxcpp, line 231: Sorry for that back-and-forth with status codes but as I mentioned [earlier](https://github.com/emersion/go-smtp/pull/116#discussion_r475574014), 500 5.5.X (0?) is probably what should be used in this particular case. 

Otherwise, LGTM.

nsmith5: Np. I'll switch that now :D

nsmith5:
@foxcpp Should be all good to go now :tada:

foxcpp:
Thanks!

PR #117, opened 2020-09-08
fix: int64 converted to string with unclear type

proletarius101:
On Golang 1.15, when executing `go test -buildmode pie -compiler gc`, it warns

```
conversion from int64 to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
```

and fails. The reason is that the type of the placeholde in formatting isn't specified. Also, explicit conversion (`string a_number`) isn't supported any longer (https://github.com/golang/go/issues/32479)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -430,7 +430,7 @@ func decodeXtext(val string) (string, error) {
 			return ""
 		}
 
-		return string(char)
+		return string(rune(char))
 	})
 	if replaceErr != nil {
 		return "", replaceErr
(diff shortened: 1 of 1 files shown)


emersion (approved):
Thanks!

emersion:
Hmm, sounds like this makes the tests fail:

```
--- FAIL: TestServer_authParam (0.00s)
    server_test.go:720: Invalid Auth value: 0xc000021770
FAIL
```

proletarius101:
Yeah, I went through this again. Fixed it now (by the trick at https://github.com/bfenetworks/bfe/pull/634/files)

foxcpp:
Thanks!

PR #120, opened 2020-09-18
patch for sim*** controllers

Snus8bit:
This server doesn't work with sim800 controller (and probably with all sim*** models) because multiline replies are sent in multiple tcp packets.  Controller ignores SMTP protocol (250- reply code) expecting that the reply is finished in one tcp packet. 
sim800 behavior:
s: 220 greets
c: EHLO me
s: 250-hello-me
s: 250-SIZE 111
s: 250-AUTH LOGIN
c: AUTH LOGIN      <- client got previous packet and began authorization
s: 250-STARTTLS    <- server keeps sending reply to EHLO
c: error, disconnect. beacuse it expected 334 VXNlcm5hbWU6
It's a totally client's bug but it can be easily fixed on the server side. Just by sending multiple lines reply in one write.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -955,14 +955,39 @@ func (c *Conn) WriteResponse(code int, enhCode EnhancedCode, text ...string) {
 		}
 	}
 
+	// Multiple tcp packets cause error
+	// in some controllers like sim800. reply must be a single tcp packet. 14.09.2020 ignatev vadim.
+	// example:
+	// s: 220 greets
+	// c: EHLO me
+	// s: 250-hello-me
+	// s: 250-SIZE 1111
+	// s: 250-AUTH LOGIN
+	// c: AUTH LOGIN      <- client got previous packet and began authorization
+	// s: 250-STARTTLS    <- server keeps sending reply to EHLO
+	// c: error, disconnect. beacuse expected 334 VXNlcm5hbWU6
+	// Essentially it is a client's problem. but...
+	reply := ""
 	for i := 0; i < len(text)-1; i++ {
-		c.text.PrintfLine("%d-%v", code, text[i])
+		reply += fmt.Sprintf("%d-%v", code, text[i])
 	}
 	if enhCode == NoEnhancedCode {
-		c.text.PrintfLine("%d %v", code, text[len(text)-1])
+		reply += fmt.Sprintf("%d %v", code, text[len(text)-1])
 	} else {
-		c.text.PrintfLine("%d %v.%v.%v %v", code, enhCode[0], enhCode[1], enhCode[2], text[len(text)-1])
+		reply += fmt.Sprintf("%d %v.%v.%v %v", code, enhCode[0], enhCode[1], enhCode[2], text[len(text)-1])
 	}
+	c.text.PrintfLine("%s", reply)
+	/*
+		for i := 0; i < len(text)-1; i++ {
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -955,14 +955,39 @@ func (c *Conn) WriteResponse(code int, enhCode EnhancedCode, text ...string) {
...
+	// Multiple tcp packets cause error
+	// in some controllers like sim800. reply must be a single tcp packet. 14.09.2020 ignatev vadim.
+	// example:
+	// s: 220 greets
+	// c: EHLO me
+	// s: 250-hello-me
+	// s: 250-SIZE 1111
+	// s: 250-AUTH LOGIN
+	// c: AUTH LOGIN      <- client got previous packet and began authorization
+	// s: 250-STARTTLS    <- server keeps sending reply to EHLO
```
foxcpp, line 967: Why client ends up assuming the list of capabilities is completed? It should wait for `250 ` (without dash). This is clearly a bug in client code and I believe go-smtp should not be monkey-patched for every single broken client implementation like that.

Though I suppose sending a capability list as a single segment might be considered a general optimization (likely an unnecessary one), in this case...

conn.go
```diff
@@ -955,14 +955,39 @@ func (c *Conn) WriteResponse(code int, enhCode EnhancedCode, text ...string) {
...
+	// c: error, disconnect. beacuse expected 334 VXNlcm5hbWU6
+	// Essentially it is a client's problem. but...
```
foxcpp, line 969: ...but?

Snus8bit: ...but  it's a simple fix. 
I totally agree with you. It's exactly why protocols were invented.
But keep in mind that those are widely used controllers. In a lot of serious industrial hardware. And by coincidence all(or most probably) of the public smtp servers like gmail are replying in a single packet. So they work fine. And this server is not working at all. It doesn't look very good from the view of the client, does it?
P.S. I wrote a claim to the manufacturer but chances on a new firmware or on a response at all are very slim.

foxcpp: > And by coincidence all(or most probably) of the public smtp servers like gmail are replying in a single packet. 

I discovered that Go's net library disables Nagle's algorithm which is enabled by default in Linux. This is likely a reason most servers except for go-smtp send responses in a single packet - you can try doing `TCPConn.SetNoDelay(false)` on a corresponding client connection to see if starts working in the same way.

conn.go
```diff
@@ -955,14 +955,29 @@ func (c *Conn) WriteResponse(code int, enhCode EnhancedCode, text ...string) {
...
+	// s: 250-SIZE 1111
+	// s: 250-AUTH LOGIN
+	// c: AUTH LOGIN      <- client got previous packet and began authorization
+	// s: 250-STARTTLS    <- server keeps sending reply to EHLO
+	// c: error, disconnect. beacuse it expected 334 VXNlcm5hbWU6
+	// Essentially it is a client's problem. but...
+	reply := ""
 	for i := 0; i < len(text)-1; i++ {
-		c.text.PrintfLine("%d-%v", code, text[i])
+		reply += fmt.Sprintf("%d-%v\r\n", code, text[i])
```
corny, line 972: Better use `string.Builder` to build the response.

corny (changes requested):

corny:
Why did you close your pull request? I would like to see this merged, because it reduces the negotiation time.

PR #122, opened 2020-09-21
Set Server Greeting to "LMTP" when using LMTP

vetinari:
as given in the example of RFC 2033 / 4.2 "The DATA command"


foxcpp:
Is there any specific reason for this change?

emersion:
Closing because clients shouldn't care.

PR #123, opened 2020-10-17
Enable LMTP over TCP

patrick246:
Changes:
 - remove check that forbids LMTP over TCP
 - introduce Network config option, to choose between TCP and unix sockets
 - add a warning to the documentation that LMTP offers no authentication and should not be hosted unprotected

### DIFF
--- a/lmtp_server_test.go
+++ b/lmtp_server_test.go
@@ -4,8 +4,13 @@ import (
 	"bufio"
 	"errors"
 	"io"
+	"math/rand"
+	"net"
+	"os"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/emersion/go-smtp"
 )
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -198,15 +207,14 @@ func (s *Server) ListenAndServe() error {
...
-
 	addr := s.Addr
-	if addr == "" {
+	if !s.LMTP && addr == "" {
 		addr = ":smtps"
+	} else if s.LMTP && addr == "" {
+		return errNoAddressSpecified
 	}
 
 	l, err := tls.Listen("tcp", addr, s.TLSConfig)
```
foxcpp, line 220: Shouldn't s.Network handling be duplicated here?

lmtp_server_test.go
```diff
@@ -195,3 +200,103 @@ func TestServer_LMTP_DuplicatedRcpt(t *testing.T) {
...
+	}
+}
+
+func TestServer_LMTP_TCP(t *testing.T) {
+	be := new(backend)
+	s := smtp.NewServer(be)
+	s.Domain = "localhost"
+	s.AllowInsecureAuth = true
+
+	testPort := rand.Int31n(65535-1024) + 1024
```
emersion, line 244: This is fragile because the port could be in use. Use port zero instead.

server.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/emersion/go-sasl"
 )
 
-var errTCPAndLMTP = errors.New("smtp: cannot start LMTP server listening on a TCP socket")
+var errNoAddressSpecified = errors.New("lmtp: no port was defined and no default port exists for this protocol")
```
emersion, line 16: s/port/address/

server.go
patrick246: Is there a use case for TLS over unix sockets

lmtp_server_test.go
patrick246: Totally agree, do you know a way to get the actually used port then? The other test code uses Serve with a custom listener, this here tests logic in ListenAndServe and can't use that

foxcpp: Actual listening port should be available via net.Listener.Addr().

server.go
foxcpp: There are other meaningful s.Network values such as tcp4 and tcp6. 
In principle, I believe library should not add any unnecessary restrictions - there is nothing fundamentally wrong in doing TLS-over-Unix sockets and one might even find use-cases for it.

server.go
```diff
@@ -172,48 +176,72 @@ func (s *Server) handleConn(c *Conn) error {
...
-// to handle requests on incoming connections.
-//
-// If s.Addr is blank and LMTP is disabled, ":smtp" is used.
-func (s *Server) ListenAndServe() error {
-	network := "tcp"
-	if s.LMTP {
+// Listen configures a listener according to the server configuration, without TLS
+// This listener can then be passed to Serve. If access to the listener is not needed, then the
+// combined function ListenAndServe can be used
+func (s *Server) Listen() (net.Listener, error) {
```
emersion, line 182: Let's not expose this

server.go
```diff
@@ -172,48 +176,72 @@ func (s *Server) handleConn(c *Conn) error {
...
+	return s.listen(false)
+}
+
+// ListenTLS configures a listener according to the server configuration, with TLS
+// This listener can then be passed to Serve. If access to the listener is not needed, then the
+// combined function ListenAndServe can be used
+func (s *Server) ListenTLS() (net.Listener, error) {
```
emersion, line 189: Ditto

server.go
```diff
@@ -172,48 +176,72 @@ func (s *Server) handleConn(c *Conn) error {
...
+	network := s.Network
+	if !s.LMTP && network == "" {
+		network = "tcp"
+	} else if s.LMTP && network == "" {
 		network = "unix"
 	}
 
 	addr := s.Addr
 	if !s.LMTP && addr == "" {
 		addr = ":smtp"
```
emersion, line 203: Should be `smtps` if TLS is enabled

server.go
patrick246: The other option would be to put the tests that test the listener setup inside the smtp package. Would that be okay?

lmtp_server_test.go
```diff
@@ -195,3 +197,94 @@ func TestServer_LMTP_DuplicatedRcpt(t *testing.T) {
...
+
+func TestServer_LMTP_NetworkDefaultBehaviorUnix(t *testing.T) {
+	be := new(backend)
+	s := smtp.NewServer(be)
+	s.Domain = "localhost"
+	s.AllowInsecureAuth = true
+
+	s.Network = ""
+	s.LMTP = true
+	s.Addr = "./testsocket"
```
emersion, line 209: We can't do that, this will clutter the current working directory.

patrick246: We could put it in os.TempDir()

emersion:
Superseded by https://github.com/emersion/go-smtp/pull/212

Issue #125, opened 2020-10-26
SendMail should make TLS mandatory


emersion:
For more control, users can handle the connection manually via `smtp.Client`.

PR #127, opened 2020-11-16
Make TLS mandatory for SendMail

emersion:
Closes: https://github.com/emersion/go-smtp/issues/125


foxcpp (approved):
Lets do it

emersion:
I'd like to add a test with a non-STARTTLS server. Also wanted to make sure coverage doesn't go down, apparently there are other tests that cover the STARTTLS case so it's fine.

Issue #131, opened 2020-12-17
2 enhancement suggestions


iredmail:
1. Methods of `Session` interface have to return `smtp.SMTPError{...}` to customize returned smtp code and message. Although it's ok to set smtp code `250` for normal exit with `SMTPError{...}`, but will the logic of handling `SMTPError` change internally in `go-smtp` someday? Should we separate a normal exit (with custom return message) and error (e.g. `SMTPResponse` and existing `SMTPError)`? Should we expose something like `conn.WriteResponse()`?
1. No way to customize the greeting text. https://github.com/emersion/go-smtp/blob/master/conn.go#L937 We'd like to customize it and display our product name.

emersion:
Just don't try customize messages, they aren't shown anywhere to the user.

iredmail:
How about the first request, easier / clearer SMTP response?

emersion:
This also sounds like customization.

iredmail:
But why doesn't allow developers to customize SMTP response?

- We may reject message and reply 554.
- Server may have some internal error and can not even save message, we reply 451 and ask sender server to try again later.

Reasonable? :)

emersion:
In both of these cases you don't reply 250, right?

iredmail:
> In both of these cases you don't reply 250, right?

Yes, and 4xx, 5xx smtp code should end the smtp session.

emersion:
If you're hitting an error, returning an `SMTPError` is fine. If you're not hitting an error, return a nil error.

iredmail:
We return 4xx/5xx on purpose. For example:

1. Access control. we don't allow this sender server to contact us.
1. Relied resource error. e.g. can not save message on disk (disk is full), SQL operation failed and return 4xx to ask sender server to retry.

For first case, it's not an "error" like "disk is full", but we still need to reject the session.

My suggestion is: Add a new struct `SMTPReply`, which is same as `SMTPError`. The word "Error" in struct name `SMTPError` is some kind of confusing, and it introduces a concern that the internal process to handle a SMTPError may be not what developers want.

foxcpp:
1. SMTPError is treated just like any other error, even if 2xx code is used in it. 
2. Customizing non-failure responses is not supported and will not be - it is just not worth the hassle.
3. I'm not strictly against customizing the initial response ("server banner").
4. Adding a special error type to terminate the connection may be a good idea.

PR #132, opened 2020-12-17
Add new field `Banner` in `Server` struct for custom smtp greeting text

iredmail:
- Add new field `Banner` in `Server` struct for custom smtp greeting text.
- Keep `Domain` field for backward compatible, but it's nowhere used. (It was only used in greeting text, which is now replaced by `Banner`)
- Use better text for `QUIT` command (`Goodnight and good luck` -> `Bye`)

See also #131 (item 2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,7 @@ func main() {
 	s := smtp.NewServer(be)
 
 	s.Addr = ":1025"
+	s.Domain = "localhost"
 	s.ReadTimeout = 10 * time.Second
 	s.WriteTimeout = 10 * time.Second
 	s.MaxMessageBytes = 1024 * 1024
--- a/README.md
+++ b/README.md
@@ -113,7 +113,6 @@ func main() {
 	s := smtp.NewServer(be)
 
 	s.Addr = ":1025"
-	s.Domain = "localhost"
 	s.ReadTimeout = 10 * time.Second
 	s.WriteTimeout = 10 * time.Second
 	s.MaxMessageBytes = 1024 * 1024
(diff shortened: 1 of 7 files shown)


iredmail:
Dear @emersion 

Could you help review and let me know your opinion? Thanks. :)

emersion:
Per the RFC, the domain name in the greeting is mandatory:

```
   Greeting       = ( "220 " (Domain / address-literal)
                  [ SP textstring ] CRLF ) /
                  ( "220-" (Domain / address-literal)
                  [ SP textstring ] CRLF
                  *( "220-" [ textstring ] CRLF )
                  "220" [ SP textstring ] CRLF )
```

Other than that, I'm not sure what purpose a custom greeting would have, so I'm not convinced this change is worth merging.

iredmail:
Dear @emersion,

Thanks for the reply.

- I will update PR and keep the domain name in greeting text.
- The custom greeting text is some kind of branding thing. :)

emersion:
>The custom greeting text is some kind of branding thing. :)

This doesn't sound very convincing to me.

iredmail:
OK. i will close this PR.
But how about change few words in SMTP reply message?

- `Speak up` -> `Invalid command`. https://github.com/emersion/go-smtp/blob/master/conn.go#L106
- `Goodnight and good luck` -> `Bye` (used by Postfix). https://github.com/emersion/go-smtp/blob/master/conn.go#L143

emersion:
Sure, fine by me.

iredmail:
> Sure, fine by me. Doing the same as Postfix sounds like a good idea.

PR #133 passed CI and waiting for approval. :)

PR #135, opened 2021-01-22
Add client and server support for SMTP DSN extension

### LINKED ISSUE #32: DSN support
https://tools.ietf.org/html/rfc3461

foxcpp:
Breaking change since it changes Client.Rcpt and Session.Rcpt signature.

Closes #32.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -29,6 +29,18 @@ const (
 	BodyBinaryMIME BodyType = "BINARYMIME"
 )
 
+type DSNReturn string
+
+const (
+	// Attach the full copy of the message to any DSN that indicates
+	// a failure. Non-failure DSNs always contain the header only.
+	ReturnFull DSNReturn = "FULL"
+
+	// Attach only header of the message to any DSN that indicates a
+	// failure.
+	ReturnHeaders DSNReturn = "HDRS"
+)
+
 // MailOptions contains custom arguments that were
 // passed as an argument to the MAIL command.
 type MailOptions struct {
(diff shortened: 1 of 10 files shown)


client.go
```diff
@@ -400,12 +411,48 @@ func (c *Client) Mail(from string, opts *MailOptions) error {
...
+// to the command. Handling of unsupported options depends on the extension.
+//
 // If server returns an error, it will be of type *SMTPError.
-func (c *Client) Rcpt(to string) error {
+func (c *Client) Rcpt(to string, opts *RcptOptions) error {
 	if err := validateLine(to); err != nil {
 		return err
 	}
-	if _, _, err := c.cmd(25, "RCPT TO:<%s>", to); err != nil {
+	cmdStr := "RCPT TO:<%s>"
```
emersion, line 422: Is it safe to just append content to `cmdStr`? The new content could have user-supplied `%v` sequences, which `fmt` will choke on.

We should probably set `cmdStr = "RCPT TO:<%s>%s"` and have a separate `extStr` string.

conn.go
```diff
@@ -478,20 +508,68 @@ func (c *Conn) handleRcpt(arg string) {
...
+			c.WriteResponse(501, EnhancedCode{5, 5, 1}, "Was expecting TO arg syntax of TO:<address>")
+			return
+		}
+	}
+	recipient := toArgs[0]
+	if recipient == "" {
+		c.WriteResponse(501, EnhancedCode{5, 5, 2}, "Was expecting MAIL arg syntax of FROM:<address>")
+		return
+	}
+	recipient = strings.Trim(recipient, "<>")
```
emersion, line 527: Can we keep the TODO about this trim being not quite what we want?

conn.go
```diff
@@ -478,20 +508,68 @@ func (c *Conn) handleRcpt(arg string) {
...
+
+	if len(toArgs) > 1 {
+		args, err := parseArgs(toArgs[1:])
+		if err != nil {
+			c.WriteResponse(501, EnhancedCode{5, 5, 4}, "Unable to parse TO ESMTP parameters")
+			return
+		}
+
+		for key, value := range args {
+			switch key {
```
emersion, line 544: Do we want to error out on unrecognized parameter?

conn.go
```diff
@@ -478,20 +508,68 @@ func (c *Conn) handleRcpt(arg string) {
...
+			case "ORCPT":
```
emersion, line 545: Can we add a `// This space is intentionally left blank` comment on the next line? IMHO it's too easy to misread this as a fall-through.

emersion:
Do you have plans to add basic test cases for these new features?

foxcpp:
> Do you have plans to add basic test cases for these new features?

Yeah, they should be added.

client.go
foxcpp: It applies to MAIL FROM too which uses similar code.

backend.go
```diff
@@ -71,7 +122,7 @@ type Session interface {
 	// Set return path for currently processed message.
 	Mail(from string, opts MailOptions) error
 	// Add recipient for currently processed message.
-	Rcpt(to string) error
+	Rcpt(to string, opts RcptOptions) error
```
kayrus, line 125: If #145 is accepted, `opts` should be switched to a pointer as well.

conn.go
kayrus: I also don't like the existing TODO, and proposed to use `mail.ParseAddress`. How is it implemented in postfix/exim mail servers? Shall we introduce a similar function?

ikedas:
Hi,

I just wrote a DSN extension also: https://github.com/emersion/go-smtp/compare/v0.18.0...ikedas:go-smtp:dsn_extension

I think my code is better in the following respects:

* It includes the fix for PR #232 (conversely, this should be merged first).
* Takes into account the case where SMTPUTF8 extension is enabled (update by RFC6533).

Should I submit a separate PR?

emersion:
Please feel free to submit a new PR, yeah.

Do note that we can still do breaking changes to our API if we feel that they are worth it.

emersion:
Superseded by #233

PR #137, opened 2021-02-18
Add SIZE extension when MaxMessageBytes not set

norguhtar:
According to [RFC 1870](https://tools.ietf.org/html/rfc1870) SIZE should always be in extension list if it is supported by the server. Just without max size. Now 

```
250-PIPELINING
250-8BITMIME
250-ENHANCEDSTATUSCODES
250-CHUNKING
250-AUTH PLAIN
250 BINARYMIME
```
Should be

```
250-Hello test
250-PIPELINING
250-8BITMIME
250-ENHANCEDSTATUSCODES
250-CHUNKING
250-AUTH PLAIN
250-BINARYMIME
250 SIZE
```


emersion (approved):
LGTM, thanks!

PR #139, opened 2021-03-22
Disable lineLimitReader when handle BDAT data

norguhtar:
We can fix this problem https://github.com/emersion/go-smtp/issues/81#issuecomment-801659120

By add checkflag to disable limit checking on handle BDAT data

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -741,12 +741,8 @@ func (c *Conn) handleBdat(arg string) {
 		}()
 	}
 
-	prevLineLimit := c.lineLimitReader.LineLimit
 	c.lineLimitReader.LineLimit = 0
-	recoverLineLimit := func() {
-		c.lineLimitReader.LineLimit = prevLineLimit
-	}
-	defer recoverLineLimit()
+
 	chunk := io.LimitReader(c.text.R, int64(size))
 	_, err = io.Copy(c.bdatPipe, chunk)
 	if err != nil {
@@ -767,6 +763,8 @@ func (c *Conn) handleBdat(arg string) {
 	c.bytesReceived += int(size)
 
 	if last {
+		c.lineLimitReader.LineLimit = c.server.MaxLineLength
+
 		c.bdatPipe.Close()
 
 		err := <-c.dataResult
--- a/conn.go
+++ b/conn.go
@@ -743,10 +743,10 @@ func (c *Conn) handleBdat(arg string) {
 
 	prevLineLimit := c.lineLimitReader.LineLimit
 	c.lineLimitReader.LineLimit = 0
-	recover := func() {
+	recoverLineLimit := func() {
 		c.lineLimitReader.LineLimit = prevLineLimit
 	}
-	defer recover()
+	defer recoverLineLimit()
 	chunk := io.LimitReader(c.text.R, int64(size))
 	_, err = io.Copy(c.bdatPipe, chunk)
 	if err != nil {
--- a/conn.go
+++ b/conn.go
@@ -739,7 +739,7 @@ func (c *Conn) handleBdat(arg string) {
 		}()
 	}
 
-	chunk := io.LimitReader(c.text.R, int64(size))
+	chunk := io.LimitReader(c.conn, int64(size))
 	_, err = io.Copy(c.bdatPipe, chunk)
 	if err != nil {
 		// Backend might return an error early using CloseWithError without consuming
--- a/conn.go
+++ b/conn.go
@@ -40,6 +40,7 @@ type Conn struct {
 	locker     sync.Mutex
 	binarymime bool
 
+	bdatReader    *lineLimitReader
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -739,6 +742,7 @@ func (c *Conn) handleBdat(arg string) {
 		}()
 	}
 
+	c.bdatReader.DisableCheck = true
```
emersion, line 745: We need to restore the check after `BDAT` has been handled.

Maybe we can just save/restore `LineLimit` instead?

```go
prevLineLimit := c.bdatReader.LineLimit
c.bdatReader.LineLimit = 0

// Handle BDAT...

c.bdatReader.LineLimit = prevLineLimit
```

Then no need to have `DisableCheck`.

norguhtar: When bdat read chunk we got curLineLength bigger than LineLimit for example LineLimit is 0 it not fix problem. May be change
```
if r.curLineLength > r.LineLimit && !r.DisableCheck {
```
to 

```
if r.curLineLength > r.LineLimit && r.LineLimit > 0 {
```

conn.go
```diff
@@ -60,15 +61,17 @@ func newConn(c net.Conn, s *Server) *Conn {
 }
 
 func (c *Conn) init() {
+	c.bdatReader = &lineLimitReader{
```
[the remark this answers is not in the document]
norguhtar: Okay

conn.go
emersion: Oh right. I guess you found another bug in our implementation, good catch!

norguhtar: Okey i changed code like this. Add restore LineLimit after handle bdat

conn.go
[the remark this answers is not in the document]
norguhtar: Renamed

conn.go
```diff
@@ -753,6 +757,7 @@ func (c *Conn) handleBdat(arg string) {
 		}
 
 		c.reset()
+		c.lineLimitReader.LineLimit = prevLineLimit
```
emersion, line 760: Instead of duplicating these statements, can we maybe use `defer` here?

```go
c.lineLimitReader.LineLimit = 0
defer func() {
	c.lineLimitReader.LineLimit = prevLineLimit
}()
```

norguhtar: I try it. But it not worked as expected.

emersion: Hm, what went wrong exactly?

norguhtar: It just don't work. If i'm add defer and first chunk handle  return error TooLongLine.

conn.go
```diff
@@ -739,6 +741,12 @@ func (c *Conn) handleBdat(arg string) {
 		}()
 	}
 
+	prevLineLimit := c.lineLimitReader.LineLimit
+	c.lineLimitReader.LineLimit = 0
+	recover := func() {
+		c.lineLimitReader.LineLimit = prevLineLimit
+	}
+	defer recover()
```
emersion, line 749: `recover` means something else for Go, it refers to handling a panic. This is confusing, because we don't handle a panic here.

Why do we need to have an intermediary variable here?

conn.go
norguhtar: > Hm, what went wrong exactly?

I found problem. This code

```
defer func() {
	c.lineLimitReader.LineLimit = prevLineLimit
}()
```
Run go routine and this can run defer and next handle at same time. I fix that with this code

```
	recover := func() {
		c.lineLimitReader.LineLimit = prevLineLimit
	}
	defer recover()
```
this fix problem same time running defer goroutine and next handleBdat

conn.go
norguhtar: fix function name.

conn.go
emersion: I don't really understand why it doesn't work without the intermediary variable. It definitely should.

norguhtar: It worked. But defer run in goroutine. This goroutine code

```
func() {
	c.lineLimitReader.LineLimit = prevLineLimit
}()

```

And some time next handleBdat running before running defer. After this when we try copy chunk

```
 io.Copy(c.bdatPipe, chunk)
```

Read return  **ErrTooLongLine**

emersion: `defer` doesn't run in a separate goroutine...

norguhtar: Okey i'm research problem deeper and create new fix. First we should't restore limit before got BDAT LAST command. It's broke In some cases and return error  ToooLongTime. For example read next BDAT command. I'm just put restore lineLimit in *last* condition and this fix problem.

emersion: Oh you're right, if the BDAT transfer is still ongoing, we shouldn't restore the limit.

emersion:
Can we also retsore the limit in case `io.Copy` fails?

norguhtar:
Add recover on io.Copy fails

emersion (approved):
LGTM, thanks for your patience!

PR #143, opened 2021-07-05
client: Add initial greeting timeout to NewClient

foxcpp:


emersion (approved):
Note: previously users could set a timeout then call `NewClient`, but now they can't anymore. Not that I think anybody was doing that TBH. If anyone wants to customize the timeouts, they could introduce a new `ClientOptions` struct.

Thanks!

PR #145, opened 2021-07-15
Change smtp.MailOptions to a pointer

kayrus:
Pass a pointer instead of the struct copy.
Will require a corresponding change in https://github.com/emersion/go-smtp-proxy

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -72,7 +72,7 @@ type Session interface {
 	AuthPlain(username, password string) error
 
 	// Set return path for currently processed message.
-	Mail(from string, opts MailOptions) error
+	Mail(from string, opts *MailOptions) error
 	// Add recipient for currently processed message.
 	Rcpt(to string) error
 	// Set currently processed message contents and send it.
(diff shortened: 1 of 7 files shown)


kayrus:
I'm also thinking of parsing the from and rcpt emails via https://pkg.go.dev/net/mail#ParseAddress and return an error, if parsing is failed. Therefore we can pass a `*mail.Address` pointer to `Mail` and `Rcpt` methods instead of the strings.

foxcpp:
> 
> 
> I'm also thinking of parsing the from and rcpt emails via https://pkg.go.dev/net/mail#ParseAddress and return an error, if parsing is failed. Therefore we can pass a `*mail.Address` pointer to `Mail` and `Rcpt` methods instead of the strings.

NACK, mail.Address is for RFC 5322 header values, not SMTP envelope. What mail.Address represents and what mail.ParseAddress would accept is invalid for SMTP.

foxcpp:
Wondering why this is not merged yet. Since we are breaking stuff all around we can probably merge this too.
@kayrus, can you rebase your PR on master?

emersion:
Acked-by me

kayrus:
rebased

foxcpp (approved):
LGTM (server example in README needs an update too but I will take care of it)

PR #146, opened 2021-07-16
server: Remove authentication-related session creation

### LINKED ISSUE #130: Get client IP address while client connects to smtp server?
Dear @emersion

First of all, thanks for the great work. 👍 

We're trying to implement an internal SMTP server, and we'd like to reject client as earlier as possible if client IP address is not whitelisted. Currently it's ok to get client IP in `Mail()` method of [`Session` interface](https://pkg.go.dev/github.com/emersion/go-smtp#Session), but we'd like to reject it even earlier. For example, when the client connects to the smtp server.

Could you share some idea of how we can achieve this? Thank you. :)

### LINKED ISSUE #136: Add a way to hook into the HELO commands
Similar to #130, a hook to the HELO commands can be useful to implement security depending on the value of the argument provided.

Examples: check the domain syntax, its existence, whether it's correctly configured to accept emails, etc.

foxcpp:
Closes #136.
Closes #130.

Here is the backend API change that addresses my concerns from #114 and also resolves two other issues.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -235,56 +235,60 @@ func (c *Conn) protocolError(code int, ec EnhancedCode, msg string) {
 
 // GREET state -> waiting for HELO
 func (c *Conn) handleGreet(enhanced bool, arg string) {
-	if !enhanced {
-		domain, err := parseHelloArgument(arg)
-		if err != nil {
-			c.WriteResponse(501, EnhancedCode{5, 5, 2}, "Domain/address argument required for HELO")
-			return
-		}
-		c.helo = domain
-
-		c.WriteResponse(250, EnhancedCode{2, 0, 0}, fmt.Sprintf("Hello %s", domain))
-	} else {
-		domain, err := parseHelloArgument(arg)
-		if err != nil {
-			c.WriteResponse(501, EnhancedCode{5, 5, 2}, "Domain/address argument required for EHLO")
-			return
-		}
-
-		c.helo = domain
-
-		caps := []string{}
-		caps = append(caps, c.server.caps...)
-		if _, isTLS := c.TLSConnectionState(); c.server.TLSConfig != nil && !isTLS {
-			caps = append(caps, "STARTTLS")
-		}
-		if c.authAllowed() {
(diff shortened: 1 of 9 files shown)


server.go
```diff
@@ -90,15 +90,13 @@ func NewServer(be Backend) *Server {
...
+					sess := conn.Session()
+					if sess == nil {
+						panic("No session when AUTH is called")
 					}
 
-					conn.SetSession(session)
-					return nil
-				})
+					return sess.AuthPlain(username, password)
+					})
```
emersion, line 99: Indentation is off here

conn.go
```diff
@@ -552,7 +541,7 @@ func (c *Conn) handleAuth(arg string) {
 				c.WriteResponse(smtpErr.Code, smtpErr.EnhancedCode, smtpErr.Message)
 				return
 			}
-			c.WriteResponse(454, EnhancedCode{4, 7, 0}, err.Error())
+			c.WriteResponse(502, EnhancedCode{5, 7, 0}, err.Error())
```
emersion, line 544: Hm, is this code more appropriate somehow?

foxcpp: Right, on a second thought I think 454 is better. Reverted the change.

server.go
foxcpp: [Fixed](https://github.com/emersion/go-smtp/compare/ec3cfb6648fb2d78cdc8907c75a60970e7f69333..03f13880e866f1351daca7d62d22936ea3d1a6c7)

conn.go
emersion: Note, I haven't actually looked at the RFC, just wanted to know the motivation for the change.

emersion (approved):
LGTM, thanks!

conn.go
foxcpp: 502 is completely unrelated to authentication whereas 454 is "temporary authentication failure". Not sure where I got 502 from.

kayrus:
https://github.com/emersion/go-smtp-proxy must be updated as well

PR #148, opened 2021-07-20
server: Expose smtp.Conn to backends

### LINKED ISSUE #147: server: Allow to access underlying net.Conn
My use-case is obtaining the full PROXY header from https://github.com/pires/go-proxyproto.

Lazy solution: Just add net.Conn into ConnectionState.
But I feel like getting rid of ConnectionState and just passing Conn to backend is 
a better idea. We then can just add Conn.Conn() method (and probably Conn.Hostname()).

foxcpp:
Here is the rough idea of how I would resolve #147.

* `*smtp.Conn` is passed to NewSession
* ConnectionState type and Conn.State is removed
* Conn.Hostname to get HELO name
* Conn.Conn to access net.Conn
* Conn.WriteResponse, Conn.ReadLine, Conn.SetSession unexported as these should not be used by backends directly
* Conn.Reject is left as is, not sure whether it should be removed/changed anyhow
* Conn.TLSConnectionState is left as is as a shortcut.

Closes #147.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ import (
 // The Backend implements SMTP server methods.
 type Backend struct{}
 
-func (bkd *Backend) NewSession(_ smtp.ConnectionState) (smtp.Session, error) {
+func (bkd *Backend) NewSession(_ *smtp.Conn) (smtp.Session, error) {
 	return &Session{}, nil
 }
 
--- a/backend.go
+++ b/backend.go
@@ -19,7 +19,7 @@ var (
 
 // A SMTP server backend.
 type Backend interface {
-	NewSession(c ConnectionState) (Session, error)
+	NewSession(c *Conn) (Session, error)
 }
 
 type BodyType string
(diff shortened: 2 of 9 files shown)


kayrus:
> Conn.WriteResponse, Conn.ReadLine, Conn.SetSession unexported as these should not be used by backends directly

@foxcpp I was hoping that you introduce a `Conn` interface, which can be extended to increase flexibility for logging, connection handling, etc.

emersion:
`smtp.Conn` can be extended even if it's not an interface.

kayrus:
Once this PR is merged, please don't hurry to make a new release. I'd like to test recent breaking changes in my project, and see how can I implement the client / server debug logs with stale or noop connections monitoring.

foxcpp:
Rebased on master to resolve conflicts

kayrus:
@foxcpp any updates on this PR?

foxcpp:
@emersion, is there any chance we can get this reviewed?

emersion:
Overall this approach LGTM. An alternative approach would be to provide something like `net/http`'s `ConnContext`, but that sounds more complicated, and can always be added later if need be.

foxcpp:
@emersion @kayrus rebased onto master

emersion:
Thanks!

PR #156, opened 2021-08-05
client: implement Dial with timeout

### LINKED ISSUE #155: DialTimeout / Incorrect address(TLS)
Hi!

# Feature request

So while using the go-smtp library, with note to `smtp.Dial` and `smtp.DialTLS` functions.
Their are no safe-guards in place that the passed `addr` is correct or even valid at all. Normally you could resolve this by adding a timeout to ensure you always get a response from the function, however this is currently not the case in `go-smtp`. Could this be added as an extra option or by de-facto default?

# Situation

Let's say I was using the alps mail server and my DNS wasn't configured right(real story).
Let's say I had specified an  `_submission` SRV instead of the correct  `_submission*s*`, alps would rightfully just call the `smtp.Dial` function instead of the correct `smtp.DialTS` function. This results in a deadlock of some sort because the Dial implementation of golang won't return any response(ever) and won't be cancelled because their are no timeouts specified.

Regards,
Gusted

Gusted:
This PR implements the `DialWithTimeout` and `DialTLSWithTimeout` functions. They work similar to their non-timeouts one, but with 1 improvement that they will timeout after the specified time.Duration.
1 Test has been added for the `DialWithTimeout` I couldn't find any tests for the normal `DIalTLS`.

- Resolves #155

### DIFF
--- a/client.go
+++ b/client.go
@@ -50,10 +50,14 @@ type Client struct {
 	DebugWriter io.Writer
 }
 
+// 30 seconds was chosen as it's the
+// same duration as http.DefaultTransport's timeout.
+var defaultTimeOut = 30 * time.Second
+
 // Dial returns a new Client connected to an SMTP server at addr.
 // The addr must include a port, as in "mail.example.com:smtp".
 func Dial(addr string) (*Client, error) {
-	conn, err := net.Dial("tcp", addr)
+	conn, err := net.DialTimeout("tcp", addr, defaultTimeOut)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


emersion:
Hm, can we make this the default instead of adding new functions?

Gusted:
That's possible, but wouldn't that be API breaking, because you would now need to provide a new parameter?

emersion:
We could have a reasonable default? e.g. `http.DefaultClient` has reasonable defaults as well. Users not happy with the default can use `NewClient` as mentioned in the issue.

Gusted:
Last time I checked `http.DefaultClient` has no timeout specified.

emersion:
`http.DefaultClient` uses `http.DefaultTransport`, which has timeouts like so: https://pkg.go.dev/net/http#DefaultTransport

emersion (approved):
LGTM, thanks!

PR #162, opened 2021-10-04
Handle tls.Conn handshake timeout

### LINKED ISSUE #161: server: simple tls.Listener can exhaust TCP resources
`ReadTimeout` and `WriteTimeout` timeouts are not reflected when a connection is `*tls.Conn`. This leads SMTPS server to wait for the TLS connection forever.

Reproducible with:
```
$ telnet localhost 465 # will hang forever
```

kayrus:
resolves #161

### DIFF
--- a/server.go
+++ b/server.go
@@ -138,6 +138,19 @@ func (s *Server) handleConn(c *Conn) error {
 		s.locker.Unlock()
 	}()
 
+	if tlsConn, ok := c.conn.(*tls.Conn); ok {
+		if d := s.ReadTimeout; d != 0 {
+			c.conn.SetReadDeadline(time.Now().Add(d))
+		}
+		if d := s.WriteTimeout; d != 0 {
+			c.conn.SetWriteDeadline(time.Now().Add(d))
+		}
+		if err := tlsConn.Handshake(); err != nil {
+			s.ErrorLog.Printf("TLS handshake error for %s: %v", tlsConn.RemoteAddr(), err)
+			return err
+		}
+	}
+
 	c.greet()
 
 	for {


emersion:
Does this happen because the handshake implicitly happens on first read or write? Would it fix it to set both read and write timeouts in `WriteResponse` and `ReadLine`, instead of just one of these?

kayrus:
@emersion no, the wait happens during the TLS handshake and it is not related to `WriteResponse` or `ReadLine`. `greet` method is blocked by TLS handshake.

kayrus:
@emersion unfortunately it is not handled. 
When I set this debug:

```
conn.go
+++ b/vendor/github.com/emersion/go-smtp/conn.go
@@ -956,6 +956,7 @@ func (c *Conn) greet() {
 func (c *Conn) WriteResponse(code int, enhCode EnhancedCode, text ...string) {
        // TODO: error handling
        if c.server.WriteTimeout != 0 {
+               fmt.Printnf("set timeout: %s\n", c.server.WriteTimeout)
                c.conn.SetWriteDeadline(time.Now().Add(c.server.WriteTimeout))
        }

````

it is triggered, but not reflected on plain TCP connection, when the listener is SMTPS:


```
set timeout: 5s
# nothing happens, connection is still open
```

See http code for an example: https://github.com/golang/go/blob/94323206aee1363471a4ae3b8d40dd4ae7a5cd9c/src/net/http/server.go#L1818-L1850

**Comment by emersion** [comment]
Try adding `SetReadDeadline` instead.

**Comment by kayrus** [comment]
@emersion I set both:
```
        smtpServer.ReadTimeout = 5 * time.Second
        smtpServer.WriteTimeout = 5 * time.Second
```

Doesn't help. Feel free to reproduce this bug in your local env.

emersion:
No, what I mean is adding a `SetReadDeadline` call in `WriteResponse`. Have you tried this?

kayrus:
@emersion yes, it works, but it doesn't expose the TLS handshake error

kayrus:
@emersion what do you suggest then? add `SetReadDeadline` into the `WriteResponse` handler? From my point of view the https://github.com/golang/go/blob/94323206aee1363471a4ae3b8d40dd4ae7a5cd9c/src/net/http/server.go#L1818-L1836 approach looks cleaner.

kayrus:
@emersion please put the io.EOF back. It causes issues on L4 level TCP monitoring.

emersion:
Nack, EOF is a TLS handshake error like any other.

TBH, I'm not even sure why we log the handshake failures in `ErrorLog`. Other connection errors aren't logged there. We should only log panics in there. PRs welcome.

Issue #159, opened 2021-09-05
Add actual example in README


MichailVestnik:
Hi! 

i copied your simple example from README and cannot use it. Official Go extension in VS Code find errors in server example.

If you can help me figure it out, I would be very grateful! 

![Снимок экрана от 2021-09-05 12-48-13](https://user-images.githubusercontent.com/87961969/132114478-f68a2fd9-2f13-43d3-a2c9-5b9b567533fa.png)
![Снимок экрана от 2021-09-05 12-48-23](https://user-images.githubusercontent.com/87961969/132114479-28ced7d9-21a3-4177-97c5-74681f9d8462.png)

Спасибо!

iredmail:
The example in README is for upcoming 0.16.0, which is not backward compatible.

MichailVestnik:
Thanks.

PR #163, opened 2021-10-09
Fix Sample Server Code To Actually Run

dweekly:
The prior code didn't run/work - with these changes there's now a "hello world" that actually runs the server.

Original Errors:
# command-line-arguments
./server.go:17:9: cannot use &Session{} (type *Session) as type smtp.Session in return argument:
	*Session does not implement smtp.Session (wrong type for Mail method)
		have Mail(string, *smtp.MailOptions) error
		want Mail(string, smtp.MailOptions) error
./server.go:58:21: cannot use be (type *Backend) as type smtp.Backend in argument to smtp.NewServer:
	*Backend does not implement smtp.Backend (missing AnonymousLogin method)

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,14 @@ func (bkd *Backend) NewSession(_ smtp.ConnectionState, _ string) (smtp.Session, 
 	return &Session{}, nil
 }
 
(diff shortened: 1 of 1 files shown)


foxcpp:
https://github.com/emersion/go-smtp/issues/159#issuecomment-913142751

@emersion i think we should tag a release already

PR #165, opened 2021-11-03
Create low level Client.SendMail utility function

petersanchez:
For cases when more control is needed than what the helper SendMail provides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,40 @@ func main() {
 }
 ```
 
-If you need more control, you can use `Client` instead.
+If you need more control, you can use `Client` instead. For example, if you
+want to send an email via a server without TLS or auth support, you can do
+something like this:
+
+```go
+package main
+
+import (
+	"log"
+	"strings"
+
+	"github.com/emersion/go-smtp"
+)
+
+func main() {
+	// Setup connection to mail server, return Client instance
+	c, err := smtp.Dial("mail.example.com:25")
+	if err != nil {
+		return err
+	}
+	defer c.Close()
+
+	// Set the sender and recipient, and send the email all in one step.
+	to := []string{"recipient@example.net"}
+	msg := strings.NewReader("To: recipient@example.net\r\n" +
+		"Subject: discount Gophers!\r\n" +
+		"\r\n" +
+		"This is the email body.\r\n")
+	err := c.SendMail("sender@example.org", to, msg)
+	if err != nil {
+		log.Fatal(err)
+	}
+}
+```
 
 ### Server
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,41 @@ func main() {
...
+
+func main() {
+	// Setup connection to mail server, return Client instance
+	c, err := smtp.Dial("mail.example.com:25")
+	if err != nil {
+		return err
+	}
+	defer c.Close()
+
+	// Connect to the server, authenticate, set the sender and recipient,
```
emersion, line 70: Remove "authenticate" here

client.go
```diff
@@ -502,6 +502,33 @@ func (c *Client) LMTPData(statusCb func(rcpt string, status *SMTPError)) (io.Wri
 	return &dataCloser{c, c.Text.DotWriter(), statusCb}, nil
 }
 
+// SendMail is the low level email delivery agent.
```
emersion, line 505: It would be nice to describe exactly what this function does, and what it doesn't do. Can take inspiration from the existing `SendMail` function. e.g. something like:

```go
// SendMail sends an email from address from, to addresses to, with message r.
//
// This function does not start TLS, nor does it perform authentication. Use
// StartTLS and Auth before-hand if desirable.
//
// The addresses in the to parameter are the SMTP RCPT addresses.
//
// The r parameter should be an RFC 822-style email with headers first, a blank
// line, and then the message body. The lines of r should be CRLF terminated.
```

client.go
```diff
@@ -456,16 +456,16 @@ func (d *dataCloser) Close() error {
...
+
+	_, _, err := d.c.Text.ReadResponse(250)
+	if err != nil {
+		if protoErr, ok := err.(*textproto.Error); ok {
+			return toSMTPErr(protoErr)
 		}
-		return nil
+		return err
 	}
+	return nil
```
emersion, line 468: Hm, why this change in `dataCloser.Close`? I don't think it's correct, I'd assume it breaks LMTP.

emersion:
When the review comments are addressed, please squash all of the commits together into a single logical commit.

client.go
petersanchez: Ah actually I meant to ask you about this. gopls was complaining about the block so I adjusted it and ran tests (passed) so I assumed it was fine. Was going to ask first and forgot. I'll revert this though.

emersion: Oh, I missed the `return` in the `if` branch above. So I guess both are the same.

I'd prefer to keep it the way it was before though, because it makes it clearer what the two cases are.

petersanchez:
@emersion All should be resolved now. Let me know if you'd like any other changes to this.

emersion (approved):
LGTM, thanks!

PR #168, opened 2021-11-26
Remove hostname argument from the NewSession method

kayrus:
Client's hostname is already available inside the `ConnectionState` struct:

https://github.com/emersion/go-smtp/blob/30169acc42e795e5d35ce901c8387950b103dfd9/conn.go#L238-L246

https://github.com/emersion/go-smtp/blob/30169acc42e795e5d35ce901c8387950b103dfd9/conn.go#L206-L213

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ import (
 // The Backend implements SMTP server methods.
 type Backend struct{}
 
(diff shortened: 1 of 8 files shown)


emersion:
LGTM

PR #169, opened 2021-11-26
Add ServerDomain member into the ConnectionState struct

kayrus:
Add server's hostname (domain)  into the `ConnectionState` struct.
It can be used for logging purposes.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -21,10 +21,11 @@ import (
 const errThreshold = 3
 
 type ConnectionState struct {
-	Hostname   string
-	LocalAddr  net.Addr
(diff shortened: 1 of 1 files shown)


emersion:
Hm, can't you just get it from the `Server` directly?

kayrus:
when you pass only the state to some internal func, there could be no server struct, only the state. introducing a new struct wrapper around that contains `ServerDomain` member looks like an overkill... :\

emersion:
`ConnectionState` has been dropped, superseded by `NewSession`.

PR #173, opened 2021-12-13
Store last SMTP code and message in Client

macrosonline:


emersion:
What's the motivation for this?

macrosonline:
> What's the motivation for this?

I need to get SMTP code when sending mails, to know "is that error temporary or not" or to get full smtp response when message is accepted (some systems shows internal message id in response and this id is required to debug delivery).

emersion:
If there is an SMTP error, the returned error will be of type `smtp.SMTPError`.

macrosonline:
> If there is an SMTP error, the returned error will be of type `smtp.SMTPError`.

But how about SMTP response message, especially if it is not error ?

emersion:
Closing as duplicate of https://github.com/emersion/go-smtp/issues/189

PR #176, opened 2021-12-16
client: add InitConn method

### LINKED ISSUE #175: SMTP client must have an ability to set greeting timeout
There is only one way to initialize an SMTP client: `NewClient` function.
This function doesn't allow to set greetings timeout and it has hardcoded 5 minutes timeout, even if the underlying connection already has a deadline, it is being overwritten:

https://github.com/emersion/go-smtp/blob/30169acc42e795e5d35ce901c8387950b103dfd9/client.go#L104-L105

If there is a network issue and TCP connection is established, but server doesn't respond with a greeting due to some reasons, the client will hang for an uncontrolled period of time.

kayrus:
`InitConn()` method allows to set the client's connection after the `smtp.Client` variable initialization with custom timeouts.

resolves #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,60 @@ func main() {
 }
 ```
 
+You can also set the timeout for the server greeting stage:
+
+```go
+package main
+
+import (
+	"log"
+	"net"
+	"strings"
+	"time"
+
+	"github.com/emersion/go-smtp"
+)
+
+var (
+	timeout = 1 * time.Second
+	addr    = "localhost:25"
+)
+
+func main() {
+	conn, err := net.DialTimeout("tcp", addr, timeout)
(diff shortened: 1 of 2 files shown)


emersion:
This doesn't sound like a great API. Instead, what about adding a `NewClientOptions` arg to `NewClient`?

kayrus:
I didn't want to break the backward compatibility and make it as much transparent as possible. I'm fine to introduce `ClientOptions` with a new function like `NewClientWithOptions` or `NewClientOpts`.

emersion:
Works for me.

emersion:
Alternative approach in https://github.com/emersion/go-smtp/pull/184

PR #177, opened 2021-12-27
Fix server example in README.md

binwiederhier:


emersion:
Nack, the example in README uses the API from master.

PichuChen:
For someone stuck in the server example on Readme.md

you can go with `go get github.com/emersion/go-smtp@master` to get master version package.

Issue #180, opened 2022-03-10
Server.Serve exits even on temporary errors


drdaeman:
Hey, thank you for this library! I was experimenting with custom net.Listeners and it seems that `Serve` exits on any error from `l.Accept` unconditionally:
https://github.com/emersion/go-smtp/blob/f9aa4f40ae7468273caf937f2742c36346de08a4/server.go#L113-L122

Unfortunately, this essentially causes server to shut down upon non-fatal errors. This potentially allows a DoS attack (shutting server down by maliciously closing connections as they're about to get accepted), although this is only my suspicion that I haven't properly validated. IIRC, TLS listener may return `net.ErrClosed` if something bad happens during the handshake, and even plain TCP connections may accidentally fail with `ECONNRESET` or `ECONNABORTED`. Either way, I believe a slightly more sophisticated loop would be an improvement.

Here's an adapted version based on [how net/http does it](https://github.com/golang/go/blob/00535b839841227ba60c2de78fbf767088f865bc/src/net/http/server.go#L3039-L3060):
```go
var tempDelay time.Duration // how long to sleep on accept failure

for {
	c, err := l.Accept()
	if err != nil {
		select {
		case <-s.done:
			// we called Close()
			return nil
		default:
		}
		if ne, ok := err.(net.Error); ok && ne.Temporary() {
			if tempDelay == 0 {
				tempDelay = 5 * time.Millisecond
			} else {
				tempDelay *= 2
			}
			if max := 1 * time.Second; tempDelay > max {
				tempDelay = max
			}
			s.ErrorLog.Printf("accept error: %v; retrying in %v", err, tempDelay)
			time.Sleep(tempDelay)
			continue
		}
		return err
	}
	go s.handleConn(newConn(c, s))
}
``` 
(Possibly somewhat contradictory, those errors are marked as "temporary" - see https://github.com/golang/go/issues/6163 for more details).

emersion:
Indeed, we shouldn't exit on temporary `Accept` errors. Patches welcome!

Issue #167, opened 2021-11-10
core: handleConn's return error cannot be catched


kayrus:
There there is no way to catch errors, returned by handleConn method:
https://github.com/emersion/go-smtp/blob/30169acc42e795e5d35ce901c8387950b103dfd9/server.go#L124

emersion:
How does `net/http` handle this?

kayrus:
@emersion `net/http`'s `serve` method doesn't return err, it passes it to the `ErrorLog` handler.

emersion:
Hm, and `ErrorLog` seems to be used for many kind of failures, both server-side panics, TLS handshakes and handler errors.

kayrus:
BTW, I noticed more limitations in `ErrorLog` handler:
* it doesn't preserve the original err, it contains only the error text (`"%v", err` instead of the `"%w", err`). 
* it doesn't preserver the TCP connection state `c.State()`

If I'd have a logging struct satisfying the `ErrorLog` interface, I'd like to have an ability to get the original error and the TCP connection state data.

PR #181, opened 2022-03-10
Improved error handling in Server.Serve

drdaeman:
Hey, not sure if you'd like the test implementation (that mock listener might be a bit weird - if you feel so, would appreciate any suggestions on how it can be improved) but here is what I came up with to address #180.

Now `Serve` does not exit on non-fatal errors from `Accept`. It also logs `handleConn` errors to the `ErrorLog` - which is not a solution but should be a partial improvement for #167.


emersion (approved):
LGTM, thanks!

PR #182, opened 2022-04-28
Add Noop method to the Session interface

kayrus:
And add error pointer to the Reset method

`Noop()` is supposed to be used to test the connection state

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -5,16 +5,17 @@ import (
 )
 
 var (
-	ErrAuthRequired    = &SMTPError{
-		Code: 502,
+	ErrAuthRequired = &SMTPError{
+		Code:         502,
 		EnhancedCode: EnhancedCode{5, 7, 0},
-		Message: "Please authenticate first",
+		Message:      "Please authenticate first",
 	}
-	ErrAuthUnsupported    = &SMTPError{
-		Code: 502,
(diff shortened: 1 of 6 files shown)


emersion:
I'm not sure I understand:

- This PR just adds a new method but never calls it.
- `NOOP` always make the server reply with `OK` immediately. The backend doesn't need to be involved in this.

emersion:
`Session` is a type used by servers only.

kayrus:
It is used by proxy. Proxy is kinda client-server.

emersion:
Sorry, I don't understand what you're trying to do, but I'm not going to merge this change, it doesn't make sense to me.

kayrus:
@emersion Your questions don't make sense :\ This PR just extends the session interface. Session interface is used in https://github.com/emersion/go-smtp-proxy. Each session represents a connection. In order to keep running long living connections I need to periodically execute NOOP command. If the NOOP fails, the connection would be reestablished.

kayrus:
@emersion I see you just archived the https://github.com/emersion/go-smtp-proxy repo. At least google cache shows that it was still active on 25th of April. Even though, I still would like to have a `Noop` method in upstream. I hope you can review the PR once again and merge it.

emersion:
I don't have any more free time to spend on your issue.

PR #184, opened 2022-05-02
client: delay greeting

emersion:
This makes it so NewClient never blocks, much like tls.Client.
This allows callers to have better control over timeouts.

### DIFF
--- a/client.go
+++ b/client.go
@@ -62,10 +62,7 @@ func Dial(addr string) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	client, err := NewClient(conn)
-	if err != nil {
-		return nil, err
-	}
+	client := NewClient(conn)
 	client.serverName, _, _ = net.SplitHostPort(addr)
 	return client, nil
 }
@@ -83,17 +80,14 @@ func DialTLS(addr string, tlsConfig *tls.Config) (*Client, error) {
 	if err != nil {
 		return nil, err
 	}
-	client, err := NewClient(conn)
-	if err != nil {
-		return nil, err
-	}
+	client := NewClient(conn)
 	client.serverName, _, _ = net.SplitHostPort(addr)
 	return client, nil
 }
 
 // NewClient returns a new Client using an existing connection and host as a
 // server name to be used when authenticating.
-func NewClient(conn net.Conn) (*Client, error) {
+func NewClient(conn net.Conn) *Client {
 	c := &Client{
 		localName: "localhost",
 		// As recommended by RFC 5321. For DATA command reply (3xx one) RFC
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -164,12 +148,30 @@ func (c *Client) Close() error {
 	return c.Text.Close()
 }
 
+func (c *Client) greet() error {
+	// Initial greeting timeout. RFC 5321 recommends 5 minutes.
+	c.conn.SetDeadline(time.Now().Add(5 * time.Minute))
```
kayrus, line 153: Let's make this configurable, e.g. `c.CommandTimeout` or `c.GreetingTimeout`.

emersion: Good idea.

client.go
```diff
@@ -164,12 +148,30 @@ func (c *Client) Close() error {
...
+	return nil
+}
+
 // hello runs a hello exchange if needed.
 func (c *Client) hello() error {
 	if !c.didHello {
 		c.didHello = true
-		err := c.ehlo()
-		if err != nil {
+		if err := c.greet(); err != nil {
```
kayrus, line 172: don't we need to run `ehlo` independently on `greet`?

kayrus: you're right. bad code styling. better to call these steps without if/then/else.
https://levelup.gitconnected.com/how-you-can-avoid-using-else-in-your-code-871197a1adbc?gi=147a42e38aca

kayrus (approved):
Generally LGTM. I have only one request to use more clean code in `hello` method without if / then if / else.

client.go
kayrus: `c.helloError` is not used anywhere. It doesn't make sense to store it as a `Client` member.
As well as `isTLS`:
```
$ grep isTLS client.go 
        _, isTLS := conn.(*tls.Conn)
        c.tls = isTLS
```

emersion: `c.helloError` is used when calling `c.hello()` repeatedly.

kayrus: > c.helloError is used when calling c.hello() repeatedly

true again. but the point with `c.tls` is still valid. it is not used anywhere.

kayrus: >> c.helloError is used when calling c.hello() repeatedly

> true again

I read the code and didn't get the point for this behavior. moreover `c.helloError` is not used, when you're calling `c.ehlo` from StartTLS method.

kayrus: Just noticed another issue. When you're using 587 port usually you prefer to introduce the client side with a safe `mail.example.com` anonymous hostname (e.g. see openssl behavior: https://github.com/openssl/openssl/blob/a3e53d56831adb60d6875297b3339a4251f735d2/apps/s_client.c#L2231). And once you upgrade the connection with StartTLS you can introduce yourself with a real hostname. Currently this behavior is impossible, because the only way to set the `c.localName` is via `Hello()` method, which returns error:
https://github.com/emersion/go-smtp/blob/608f3c2840584931f2867716078c43485ea7ef3f/client.go#L186-L195

emersion: All of these are unrelated to this PR.

emersion: `Client.tls` dropped in 6ecbad7e0eb0.

emersion: Opened https://github.com/emersion/go-smtp/issues/245 for the hostname thing.

PR #188, opened 2022-06-03
Handle custom responses for Mail and Rcpt methods

### LINKED ISSUE #187: Server: handle a custom response for "Mail" and "Rcpt" session methods
I can return a custom `*SMTPError` object, but `handleMail` and `handleRcpt` methods won't set the `c.fromReceived` or add the recipient into the `c.recipients`.

See below:

https://github.com/emersion/go-smtp/blob/608f3c2840584931f2867716078c43485ea7ef3f/conn.go#L400-L410

https://github.com/emersion/go-smtp/blob/608f3c2840584931f2867716078c43485ea7ef3f/conn.go#L488-L497

I suggest the following fix:

```patch
--- a/conn.go
+++ b/conn.go
@@ -419,6 +419,9 @@ func (c *Conn) handleMail(arg string) {
        if err := c.Session().Mail(from, opts); err != nil {
                if smtpErr, ok := err.(*SMTPError); ok {
                        c.WriteResponse(smtpErr.Code, smtpErr.EnhancedCode, smtpErr.Message)
+                       if smtpErr.Code == 250 {
+                               c.fromReceived = true
+                       }
                        return
                }
                c.WriteResponse(451, EnhancedCode{4, 0, 0}, err.Error())
@@ -507,6 +510,9 @@ func (c *Conn) handleRcpt(arg string) {
        if err := c.Session().Rcpt(recipient); err != nil {
                if smtpErr, ok := err.(*SMTPError); ok {
                        c.WriteResponse(smtpErr.Code, smtpErr.EnhancedCode, smtpErr.Message)
+                       if smtpErr.Code == 250 {
+                               c.recipients = append(c.recipients, recipient)
+                       }
                        return
                }
                c.WriteResponse(451, EnhancedCode{4, 0, 0}, err.Error())
````

kayrus:
Resolves #187

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -400,6 +400,9 @@ func (c *Conn) handleMail(arg string) {
 	if err := c.Session().Mail(from, opts); err != nil {
 		if smtpErr, ok := err.(*SMTPError); ok {
 			c.WriteResponse(smtpErr.Code, smtpErr.EnhancedCode, smtpErr.Message)
+			if smtpErr.Code == 250 {
+				c.fromReceived = true
+			}
 			return
 		}
 		c.WriteResponse(451, EnhancedCode{4, 0, 0}, err.Error())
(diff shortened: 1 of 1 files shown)


kayrus:
@emersion kindly ping

kayrus:
@emersion kindly ping

emersion:
I'm not a fan of returning an error when there is no actual error.

emersion:
Closing, I don't think this is the right approach here.

PR #190, opened 2022-06-03
Read server response on DATA command

### LINKED ISSUE #189: Client: read server response on DATA command
Usually remote SMTP servers return a successful 250 code with an informative message, e.g. `Queued %messageID%`
Unfortunately current implementation doesn't allow to intercept this response, though it's required for message tracking and monitoring.

https://github.com/emersion/go-smtp/blob/608f3c2840584931f2867716078c43485ea7ef3f/client.go#L460-L467

kayrus:
This is a braking change. Now `Data()` method will accept a callaback function and `LMTPData()` method callbacks will receive success statuses instead nil. Therefore dependent `LMTPData` code must check for status response code.

Resolves #189 

P.S. I rewrote the `(d *dataCloser) Close()` code to make it more readable.

### DIFF
--- a/client.go
+++ b/client.go
@@ -429,7 +429,8 @@ func (c *Client) Rcpt(to string) error {
 type dataCloser struct {
 	c *Client
 	io.WriteCloser
-	statusCb func(rcpt string, status *SMTPError)
+	statusCb     func(status *SMTPError)
+	lmtpStatusCb func(rcpt string, status *SMTPError)
 }
 
 func (d *dataCloser) Close() error {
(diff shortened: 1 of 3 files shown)


Slugger:
I came here with the exact same dilemma -- I need to capture & log the  `250` response to all `DATA` commands.  I was about to write a PR myself then saw this. 😄 

The semantics of returning an `SMTPError` on a 250 response seems a little off to me since this isn't really an error condition.  My thought on this was to add some kind of additional property to the client struct that was a string called `DataResponse` or something like that.  It just holds the response to the last `DATA` command issued (when not an error, errors would still be returned as usual).  This would also avoid the breaking change to `Close()` as discussed.

But I'll gladly take this as well.  Any indication of whether this is going to be accepted?

kayrus:
@Slugger then I can simply add the same functionality as a callback function in LMTP session.

kayrus:
@Slugger @emersion I updated the PR, let me know your feedback.

Slugger:
I like this approach better for sure.  Now I can handle the `DATA` response as needed via a callback.

emersion:
Sorry for the delay. I think in general I'm not a huge fan of callbacks... Could we instead turn the returned `io.WriteCloser` into something like a `*DataCommand`, and expose the info message as a method?

foxcpp:
Agreed with emersion, I think adding an explicit method looks like better design to me.

I am also not sure if SMTPError is the right type to use here, I would at least create an alias named SMTPResponse or something - some careful refactoring may be needed here.

emersion:
Do we really need to expose the full response? Is there anything interesting in there apart from the human-readable message?

Slugger:
Usually a message ID from the upstream receiver.  My use case at the time was that I was trying to use this as a piece in an email forwarding service and when I forwarded messages upstream I wanted to capture the message id from the receiver for logging, tracking and auditing purposes.

emersion:
Closing in favor of https://github.com/emersion/go-smtp/pull/263

PR #193, opened 2022-07-03
feat(): add MaxConcurrentConnections

n33pm:


emersion:
NACK, sorry. This can be implemented with your own `net.Listener`.

PR #200, opened 2022-10-08
Add flag to wait for conns to close on Server.Close()

GrahamTonySmith:
Add a waitgroup to Server struct.

Add a SoftClose bool flag to Server struct.

Extend the Serve function receiver on the Server struct to increment waitgroup before handling a connection and decrement the waitgroup when the handling is done.

Extend the Close function receiver on the Server struct to wait for all connection handling to return before returning from close if Server.SoftClose is true.

### DIFF
--- a/server.go
+++ b/server.go
@@ -44,6 +44,7 @@ type Server struct {
 	ErrorLog          Logger
 	ReadTimeout       time.Duration
 	WriteTimeout      time.Duration
+	SoftClose         bool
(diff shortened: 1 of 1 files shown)


emersion:
Can we instead add a separate `Shutdown` function, just like `net/http.Server`?

GrahamTonySmith:
Absolutely, great idea.

PR #202, opened 2022-10-21
Fixes for example README server code

jasonfriedland:
Fixes for example README server code to compile successfully, remove warnings etc.

- Add missing interface methods on `Backend` implementation, fix pointer arg on `Mail()`.
- `ioutl` is deprecated as of Go 1.16
- Minor style (error capitalisation, receiver vars).

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,31 +88,34 @@ package main
 import (
 	"errors"
 	"io"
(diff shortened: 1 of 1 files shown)


emersion:
The README has example code for the next go-smtp version.

Issue #26, opened 2019-01-21
Send message with ssl/tls


djeckson:
How to send a message using ssl/tls?

emersion:
https://godoc.org/github.com/emersion/go-smtp#DialTLS

djeckson:
How to send a message using smtp client?

emersion:
You can have a look at `SendMail`'s source code.

PR #204, opened 2022-10-24
client: SendMailTLS

d2ro:
[RFC 8314](https://www.rfc-editor.org/rfc/rfc8314) recommends that:

> Connections to Mail Submission Servers and Mail Access Servers be made using "Implicit TLS" (as defined below), in preference to connecting to the "cleartext" port and negotiating TLS using the STARTTLS command or a similar command.

go-smtp has `Dial`, `DialTLS`, and `SendMail` functions. Adding `SendMailTLS` would add implicit TLS (aka SMTPS) support in a consistent way.

This is related to #26. I have tried to put `SendMailTLS` into my code, but that is not feasible because it accesses some unexported struct fields and functions.

`SendMailTLS` is basically `SendMail`, but with `DialTLS` instead of `Dial` and without the StartTLS stuff.

### DIFF
--- a/client.go
+++ b/client.go
@@ -608,6 +608,36 @@ func SendMail(addr string, a sasl.Client, from string, to []string, r io.Reader)
 	return c.SendMail(from, to, r)
 }
 
+// SendMailTLS works like SendMail, but with implicit TLS.
+func SendMailTLS(addr string, a sasl.Client, from string, to []string, r io.Reader) error {
+	if err := validateLine(from); err != nil {
+		return err
+	}
+	for _, recp := range to {
+		if err := validateLine(recp); err != nil {
+			return err
+		}
+	}
+	c, err := DialTLS(addr, nil)
+	if err != nil {
+		return err
+	}
+	defer c.Close()
+
+	if err = c.hello(); err != nil {
+		return err
+	}
+	if a != nil && c.ext != nil {
+		if _, ok := c.ext["AUTH"]; !ok {
+			return errors.New("smtp: server doesn't support AUTH")
+		}
+		if err = c.Auth(a); err != nil {
+			return err
+		}
+	}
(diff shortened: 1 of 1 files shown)


emersion:
> I have tried to put SendMailTLS into my code, but that is not feasible because it accesses some unexported struct fields and functions.

It should be possible to use just the exported API for this. The `hello` call can be replaced with `Hello`, for instance.

Regardless, it may be a good idea to have this function.

d2ro:
Using the external API with the latest release v0.15.0 would look like this. (The diff is based on a copy of `func SendMail`).

```diff 
+// validateLine checks to see if a line has CR or LF as per RFC 5321
+func validateLine(line string) error {
+       if strings.ContainsAny(line, "\n\r") {
+               return errors.New("smtp: A line must not contain CR or LF")
+       }
+       return nil
+}
+
-func SendMail(addr string, a sasl.Client, from string, to []string, r io.Reader) error {
+func SendMailTLS(addr string, a sasl.Client, from string, to []string, r io.Reader) error {
```

```diff
-       c, err := Dial(addr)
+       c, err := smtp.DialTLS(addr, nil)
        if err != nil {
                return err
        }
        defer c.Close()
-       if err = c.hello(); err != nil {
-               return err
-       }
-       if ok, _ := c.Extension("STARTTLS"); ok {
-               if err = c.StartTLS(nil); err != nil {
-                       return err
-               }
-       }
-       if a != nil && c.ext != nil {
-               if _, ok := c.ext["AUTH"]; !ok {
+       supportsAuth, _ := c.Extension("AUTH") // calls c.hello()
+       if a != nil {
+               if !supportsAuth {
                        return errors.New("smtp: server doesn't support AUTH")
                }
```

foxcpp (approved):

emersion:
Thanks!

PR #210, opened 2023-02-01
support AUTH LOGIN

firefart:
this PR adds support for the AUTH LOGIN mechanism


emersion:
Thanks for your PR, but this is something we've dropped because LOGIN is non-standard.

firefart:
@emersion thanks. Would it be possible to add a flag on server instantion to allow login auths?

emersion:
This is not something I want to support upstream. You can implement the interfaces in your own code though.

PR #212, opened 2023-02-24
Both SMTP and LMTP servers may listen either TCP or UNIX domain socket

ikedas:
*   [RFC2033](https://datatracker.ietf.org/doc/rfc2033/) does _not_ state that "LMTP mode cannot be used with a TCP listener": It states that it "_MUST NOT be used on the TCP **service port 25**_".  In fact, most LMTP implementations support TCP, and the default port number is often 24.
*   There is at least one implementation that supports SMTP over UNIX domain socket: [OpenSMTPD](https://man.openbsd.org/smtpd.conf.5).
*    TLS over UNIX domain socket seems worthwhile. On some systems, since permissions of pathname socket  are ignored, they cannot be relied upon (See [unix(7)](https://man7.org/linux/man-pages/man7/unix.7.html)).

That's why this PR makes changes to allow both SMTP and LMTP servers to be used on either TCP or UNIX domain socket with optional TLS.

---
Some other changes:
*   <del>The server address is treated as UNIX domain socket when the value of `Server.Addr` contains "/".</del>
* <ins>New field `Network` for the type of connection, defaulting `"tcp"`, is introduced. `LMTP` field no longer affects the type.</ins>  
*   The default address for `ListenAndServeTLS()` was changed to "`:465`" because on several systems "`:smtps`" cannot be resolved.

### DIFF
--- a/server.go
+++ b/server.go
@@ -15,7 +15,6 @@ import (
 )
 
 var (
-	errTCPAndLMTP   = errors.New("smtp: cannot start LMTP server listening on a TCP socket")
 	ErrServerClosed = errors.New("smtp: server already closed")
 )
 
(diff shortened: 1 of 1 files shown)


ikedas:
I noticed that a part of the goals of this PR is duplicated in the PR #123 by @patrick246 . However I believe my changes are better.

emersion:
I am not a fan of making `Server.Addr` handling more complicated.

ikedas:
> I am not a fan of making `Server.Addr` handling more complicated.

Is another option like `Server.Network` preferred?  I don't care either way.

ikedas:
@emersion , I removed auto-detection of network type.
If I should squash the commits, let me know.

emersion:
Changed the logic a bit, thanks!

PR #213, opened 2023-03-10
server: Correct status codes in several responses.

### LINKED ISSUE #195: Should 'Too many recipients' return 452 4.5.3 instead of 552 5.5.3?
I've been looking into this response a bit for something I'm working on. The current response when we have a recipient limit is

```
552 5.5.3 Maximum limit of %v recipients reached
```

From my reading though, it seems the following is the more 'standard' response for this case

```
452 4.5.3 Too many recipients
```

with some providers including the limit there. Specifically I'm concerned with how the big players say to handle this status code to make sure that they send correctly with fewer recipients. I'm about to do testing with the current code and will report my findings, but I think this should be changed.

**Gmail**: [https://support.google.com/a/answer/3726730](https://support.google.com/a/answer/3726730)
> 452, "4.5.3", Your message has too many recipients. For more information regarding Google's sending limits, visit [Limits for sending & getting mail.](https://support.google.com/mail/answer/22839)

**Microsoft**: [https://support.google.com/a/answer/3726730](https://support.google.com/a/answer/3726730)
> For inbound email, the Receive connector MaxRecipientsPerMessage is verified first. However, if the number of recipients exceeds the limit, the message is not rejected; the connection receives the error, 452 4.5.3 Too many recipients. Most mail servers understand this error and they will continue to resend the message in another connection until the message is delivered to all recipients.

Granted they are talking about their own servers and how they expect clients to handle it, so I would hope they respond the same way. The rfc's may suggest that 5xx is more correct (I don't think they specifically say what to do for too many recipients), but if gmail and ms agree on something else, I would consider it the de-facto standard.

ikedas:
- The excess number of recipients should be responded with `4XX`.
- Disconnections due to network errors or idle timeouts should be `4XX`.
- `3XX` codes do not have enhanced codes.
- Other typos.
---
This PR may fix #195 .

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -924,7 +924,11 @@ func (c *Conn) Reject() {
 }
 
 func (c *Conn) greet() {
-	c.writeResponse(220, NoEnhancedCode, fmt.Sprintf("%v ESMTP Service Ready", c.server.Domain))
+	protocol := "ESMTP"
+	if c.server.LMTP {
+		protocol = "LMTP"
+	}
+	c.writeResponse(220, NoEnhancedCode, fmt.Sprintf("%v %s Service Ready", c.server.Domain, protocol))
 }
 
 func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
--- a/conn.go
+++ b/conn.go
@@ -281,7 +281,7 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 // READY state -> waiting for MAIL
 func (c *Conn) handleMail(arg string) {
 	if c.helo == "" {
-		c.writeResponse(502, EnhancedCode{2, 5, 1}, "Please introduce yourself first.")
+		c.writeResponse(502, EnhancedCode{5, 5, 1}, "Please introduce yourself first.")
 		return
 	}
 	if c.bdatPipe != nil {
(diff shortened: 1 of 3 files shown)


ikedas:
This PR may fix #195

conn.go
```diff
@@ -467,7 +467,7 @@ func (c *Conn) handleRcpt(arg string) {
 	recipient := strings.Trim(arg[3:], "<> ")
 
 	if c.server.MaxRecipients > 0 && len(c.recipients) >= c.server.MaxRecipients {
-		c.writeResponse(552, EnhancedCode{5, 5, 3}, fmt.Sprintf("Maximum limit of %v recipients reached", c.server.MaxRecipients))
+		c.writeResponse(452, EnhancedCode{4, 5, 3}, fmt.Sprintf("Maximum limit of %v recipients reached", c.server.MaxRecipients))
```
emersion, line 470: It seems like the only valid short code for x.5.3 is 451, according to https://www.iana.org/assignments/smtp-enhanced-status-codes/smtp-enhanced-status-codes.xhtml

emersion (changes requested):
LGTM apart from this

(Would've been nice to split the greeting message changes into its own commit.)

conn.go
ikedas: I found that [RCF 5321](https://www.rfc-editor.org/rfc/rfc5321#section-4.5.3.1.9) states that basic status code for "Too many recipients" is `452`.  The "Associated Basic Status Code" field in IANA registry does not cover all possible values,

ikedas:
> (Would've been nice to split the greeting message changes into its own commit.)

Makes sense.  I've done.

conn.go
emersion: Right!

emersion (approved):
Thanks!

PR #214, opened 2023-03-13
Add Shutdown to Server struct

grahamtonysmith:
Add a function, server.Shutdown, which will gracefully shut down the server without interrupting any active connections (like http server).

### DIFF
--- a/server.go
+++ b/server.go
@@ -263,7 +263,7 @@ func (s *Server) ListenAndServeTLS() error {
 func (s *Server) Close() error {
 	select {
 	case <-s.done:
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -274,6 +284,45 @@ func (s *Server) Close() error {
...
+	select {
+	case <-s.done:
+		return ErrServerClosed
+	default:
+		close(s.done)
+	}
+
+	var err error
+	s.locker.Lock()
+	log.Println("listners", len(s.listeners))
```
emersion, line 304: Remnant from debugging?

PR #218, opened 2023-03-29
improve the MAIL From parser 

mabels:
improve the MAIL From parser so that common addresses like
<Test Email <test.email@test.mail>> BODY=8BITMIME
are processable.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -278,6 +278,54 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 	c.writeResponse(250, NoEnhancedCode, args...)
 }
 
+func findSharpBrackets(arg string) (string, string, error) {
+	if arg[0] != '<' {
+		return "", "", errors.New("must start with <")
+	}
+	for i := 1; i < len(arg); i++ {
+		if arg[i] == '>' {
+			return arg[0 : i+1], arg[i+1:], nil
+		}
+		if arg[i] == '<' {
+			sharp, _, err := findSharpBrackets(arg[i:])
+			if err != nil {
+				return "", "", err
+			}
+			// -1 because the for loop will increment i
+			i += len(sharp) - 1
+		}
+	}
+	return "", "", errors.New("must end with >")
+}
+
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -278,6 +278,54 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
...
+			// -1 because the for loop will increment i
+			i += len(sharp) - 1
+		}
+	}
+	return "", "", errors.New("must end with >")
+}
+
+func getFromAndArgs(arg string) (from string, args []string, err error) {
+	arg = strings.TrimSpace(arg)
+	args = strings.Split(arg, " ")
```
emersion, line 303: This line can be removed, it's unused in the `if` block and is already in the `else` block.

conn.go
```diff
@@ -278,6 +278,54 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
...
+
+	if arg[0] == '<' {
```
emersion, line 305: We should check `len(arg) > 0` before accessing the first character.

emersion:
Hm, it doesn't seem like this handles `Quoted-string` in the SMTP grammar.

mabels:
Hi,

thx I did not looked in the protocol specs --- will do try to update to a
compliment implemention

meno

mabels:
@> Hm, it doesn't seem like this handles `Quoted-string` in the SMTP grammar.

Hi,

i'm very bad in reading these:
https://www.rfc-editor.org/rfc/rfc5321#section-4.1.2
but as i try to understand in the wild the from is different from the specs.
I typical see <First Second ThirdName <bla.blub@foo.domain>> as From
So to improve i think we should either understand(find a different) rfc or look in a different smtp implemtation.

any thoughts?

mabels:
I tried to join the RFC and living parser from postfix.

emersion:
Superseded by https://github.com/emersion/go-smtp/commit/3ecf12d3a9e3654b6be32d0bbd5e0f5228601d34

PR #220, opened 2023-03-31
Improve testing

mabels:
The current testing uses this internal:
testHookStartTLS
which is not external available and i don't have control of internal structures like:
- tls.Config
- net.Listener

So i added to SendMail and ListenAndServe an optional argument which exposes or enables the modification of the tls.Config.
With this both hooks its possible to build external tests or make use of dynamic binded servers or inject an request specific tls.Config in Sendmail.

### DIFF
--- a/client.go
+++ b/client.go
@@ -6,6 +6,7 @@ package smtp
 
 import (
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/base64"
 	"errors"
 	"fmt"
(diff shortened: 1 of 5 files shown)


emersion:
This API is by design. If you want to have control over the `tls.Config`, you need to use lower-level functions.

mabels:
I known thats why i made it optional so it's not a breaking change.
If the api should be stay like it's I would recommend to make helper function which enable this.
The ListenAndServe could named ListenAndServerReady
The SendMail could named SendMailWithOptions

But to drop them could cause that we loose we might see alot more client implementation where very custom implementation of lowlevel api calls might contain an other bug.

I get your point but i want to SendMail so there should be one method to do that. And on the ListenAndServe i don't want to think about whats happens behind the scenes.

emersion:
Sorry, I don't believe adding new API for this is worth it. I think it's easy enough for library users to pass their own `net.Listener`/`net.Conn`.

PR #223, opened 2023-04-08
server: Drop root privileges

ikedas:
On many systems listening well-known ports needs root privileges.  This PR allows the server to drop privileges if it can listen on the port.  Also, this allows to set the mode of the UNIX domain socket file.

The new fields `Username`, `Groupname` and `SocketMode` are introduced to `smtp.Server`.

---
This PR may conflict with PR #212.

### DIFF
--- a/server.go
+++ b/server.go
@@ -8,7 +8,10 @@ import (
 	"log"
 	"net"
 	"os"
+	"os/user"
+	"strconv"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/emersion/go-sasl"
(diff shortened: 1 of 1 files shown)


ikedas:
Note that, with Linux, Go 1.16 or later is required, as earlier releases haven't implemented `syscall.Set[ug]id()`.

emersion:
Sorry, I would rather not have this in the library:

- This is too opinionated for a library.
- This is platform-specific. On Linux, the bind_service capability is a better way to achieve the same goal.
- This affects the global program state.
- This doesn't allow one to design servers which can be reloaded (changing the listeners after the server is started).

ikedas:
I understand. I'll withdraw this PR.

BTW I'm planning another PR adding `netutil.LimitListener()` to avoid congestion.  Should I also forgo this?

emersion:
As a rule of thumb, if `net/http` doesn't do it, go-smtp probably shouldn't do it either. Library users can easily drop privileges or use a `LimitListener` if they want to.

ikedas:
I see. @emersion , thank you for advice.  This PR is withdrawn.

PR #225, opened 2023-07-07
Update server example to get errors from smtp

matthewp:
Also there hasn't been a release since `ErrAuthFailed` was added. Mind doing a tag for that?


emersion (approved):
Thanks!

PR #226, opened 2023-07-14
Update MAIL FROM AUTH handling to accept mailbox without angle brackets

bobobo1618:
I have a client that sends a message to an SMTP server using this library (the Protonmail bridge) and encounter problems because the client doesn't enclose the `AUTH=` parameter in angle brackets.

I looked at [the relevant RFC](https://datatracker.ietf.org/doc/html/rfc4954#section-5) and it doesn't appear that this parameter requires angle brackets. [RFC2821 (referenced by the other RFC) section 4.1.2](https://datatracker.ietf.org/doc/html/rfc2821#section-4.1.2) requires angle brackets for a _Path_ but not for a Mailbox. It explicitly allows Mailboxes without angle brackets:

```
Mailbox = Local-part "@" Domain
Local-part = Dot-string / Quoted-string
            ; MAY be case-sensitive
Dot-string = Atom *("." Atom)
Atom = 1*atext
(From https://datatracker.ietf.org/doc/html/rfc2822#section-3.2.4)
atext           =       ALPHA / DIGIT / ; Any character except controls,
                        "!" / "#" /     ;  SP, and specials.
                        "$" / "%" /     ;  Used for atoms
                        "&" / "'" /
                        "*" / "+" /
                        "-" / "/" /
                        "=" / "?" /
                        "^" / "_" /
                        "`" / "{" /
                        "|" / "}" /
                        "~"
```

In other words, the `Local-part` can be composed of a `Dot-string`, which is composed only alphanumerics plus some special characters (the list of which excludes angle brackets).

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -362,20 +362,15 @@ func (c *Conn) handleMail(arg string) {
 				opts.Body = BodyType(value)
 			case "AUTH":
 				value, err := decodeXtext(value)
-				if err != nil {
+				if err != nil || value == "" {
 					c.writeResponse(500, EnhancedCode{5, 5, 4}, "Malformed AUTH parameter value")
 					return
 				}
-				if !strings.HasPrefix(value, "<") {
-					c.writeResponse(500, EnhancedCode{5, 5, 4}, "Missing opening angle bracket")
-					return
+				if value == "<>" {
+					value = ""
 				}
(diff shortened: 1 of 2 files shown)


emersion:
My reading of the spec is that angle brackets are not allowed here, apart from the special case of `<>`. I've edited the PR accordingly.

PR #230, opened 2023-08-14
Add Github workflow for automated tests

corny:


emersion:
Thanks, but we already have builds.sr.ht for CI.

corny:
There's also a possible data race:
https://github.com/digineo/go-smtp/actions/runs/5855132039/job/15872333833

emersion:
Ah, it sounds like the build jobs were marked as "unlisted". I've fixed that.

The coverage is not computed in CI right now.

Yeah, I've seen the data race locally -- sounds like a test needs fixing.

PR #232, opened 2023-08-28
Fix broken encodeXtext()

ikedas:
With this method call:
```go
email := "e=mc2@example.com"
c.Mail(email, &MailOptions{Auth: &email})
```
this should be sent to the server:
```
MAIL FROM:<e=mc2@example.com> AUTH=e+3Dmc2@example.com
```

However, following thing is sent:
```
MAIL FROM:<e=mc2@example.com> AUTH=e+65+3D=+3Dm+6Dc+632+32@+40e+65x+78a+61m+6Dp+70l+6Ce+65.+2Ec+63o+6Fm+6D
```
This PR will fix it.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -426,16 +426,14 @@ func encodeXtext(raw string) string {
 	out.Grow(len(raw))
 
 	for _, ch := range raw {
-		if ch == '+' || ch == '=' {
+		switch {
+		case ch >= '!' && ch <= '~' && ch != '+' && ch != '=':
+			// printable non-space US-ASCII except '+' and '='
+			out.WriteRune(ch)
+		default:
 			out.WriteRune('+')
 			out.WriteString(strings.ToUpper(strconv.FormatInt(int64(ch), 16)))
 		}
-		if ch > '!' && ch < '~' { // printable non-space US-ASCII
-			out.WriteRune(ch)
-		}
-		// Non-ASCII.
-		out.WriteRune('+')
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -426,16 +426,14 @@ func encodeXtext(raw string) string {
 	out.Grow(len(raw))
 
 	for _, ch := range raw {
-		if ch == '+' || ch == '=' {
+		switch {
+		case ch > '!' && ch < '~' && ch != '+' && ch != '=':
```
emersion, line 430: The RFC says:

```
      xchar = any ASCII CHAR between "!" (33) and "~" (126) inclusive,
              except for "+" and "=".
```

Shouldn't we use `ch >= '!' && ch <= '~'`  here?

emersion:
Overall LGTM, the old code was definitely broken.

conn.go
ikedas: Wow. That's a terrible mistake. I'll fix and re-push it.

emersion: It's not _that_ bad, it just means we will unnecessarily use xtext encoding for these characters as well. But yeah, it's better if we only use xtext encoding when necessary.

ikedas: However any less tolerant implementations might not accept sequences like `+21`.

ikedas: Fixed.

emersion (approved):
Thanks for the fix!

PR #233, opened 2023-09-10
Server & Client: Implementing DSN extension (RFC 3461, RFC 6533)

### LINKED ISSUE #32: DSN support

ikedas:
This may fix #32.


ikedas:
Some notes:

*   This PR depends on #232.
*   If invalid parameter values found in MAIL or RCPT command, reply-code `501` is issued (See RFC3641, 5.1), though for the other parameters currently `500` is issued.
*   Test coverage is not adequate.

client.go
```diff
@@ -400,13 +401,32 @@ func (c *Client) Mail(from string, opts *MailOptions) error {
 			return errors.New("smtp: server does not support SMTPUTF8")
 		}
 	}
+	if _, ok := c.ext["DSN"]; ok && opts != nil {
+		switch opts.Return {
+		case ReturnFull, ReturnHeaders:
+			cmdStr += "%s"
```
emersion, line 407: I'm not a fan of using params and %-formatting here. Can we use a `strings.Builder` to build the command string, and not use the %-formatting params at all? (IOW: `c.cmd(25, "%s", sb.String())`).

conn.go
```diff
@@ -461,17 +660,76 @@ func (c *Conn) handleRcpt(arg string) {
...
+				c.writeResponse(504, EnhancedCode{5, 5, 4}, "NOTIFY is not implemented")
+				return
+			}
+			notify := []DSNNotify{}
+			seen := map[DSNNotify]struct{}{}
+			for _, val := range strings.Split(value, ",") {
+				val = strings.ToUpper(val)
+				switch DSNNotify(val) {
+				case NotifyNever:
+					if len(seen) != 0 {
```
emersion, line 690: Maybe we can move this check at the end of the loop, to avoid duplicating them?

```go
for _, val := range strings.Split(value, ",") {
    …
}
if _, ok := seen[NotifyNever]; ok && len(seen) > 1 {
    // error
}
```

client.go
```diff
@@ -422,7 +442,59 @@ func (c *Client) Rcpt(to string, opts *RcptOptions) error {
...
 		return err
 	}
-	if _, _, err := c.cmd(25, "RCPT TO:<%s>", to); err != nil {
+	cmdStr := "RCPT TO:<%s>"
+	params := []interface{}{to}
+	if _, ok := c.ext["DSN"]; ok && opts != nil {
+		if opts.Notify != nil && len(opts.Notify) != 0 {
+			val := " NOTIFY="
+			seen := map[DSNNotify]struct{}{}
+			for i, v := range opts.Notify {
```
emersion, line 451: Since the server also performs the same checks, maybe worth extracting into a helper function?

client.go
ikedas: Thanks for suggestion.  Done!

conn.go
ikedas: Done!

client.go
ikedas: Done!  I'd like to squash additional commits later.

ikedas:
I've done what I could.

smtp.go
```diff
@@ -52,5 +66,29 @@ type MailOptions struct {
...
+	NotifyFailure DSNNotify = "FAILURE"
+	NotifySuccess DSNNotify = "SUCCESS"
+)
+
+type DSNAddressType string
+
+const (
+	AddressTypeRFC822 DSNAddressType = "RFC822"
+	AddressTypeUTF8   DSNAddressType = "UTF-8"
+)
```
emersion, line 83: It seems a bit inconsistent to have the types prefixed with "DSN" but not the constants. Is this deliberate? Should we prefix everything with "DSN", or do you think we can drop the "DSN" prefix from type names?

(Sorry for the delay!)

conn.go
```diff
@@ -421,6 +449,122 @@ func decodeXtext(val string) (string, error) {
 	return decoded, nil
 }
 
+// This regexp matches 'EmbeddedUnicodeChar' token defined in
+// https://datatracker.ietf.org/doc/html/rfc6533.html#section-3
+// however it is intentionally relaxed by requiring only '\x{HEX}' to be
+// present.  It also matches disallowed characters in QCHAR and QUCHAR defined
+// in above.
+// So it allows us to detect malformed values and report them appropriately.
+var eUOrDCharRe = regexp.MustCompile(`\\x[{][0-9A-F]+[}]|[[:cntrl:] \\+=]`)
```
emersion, line 458: In general I'm not a fan of using regexp for decoding but we already do that in go-smtp so it's not a big deal…

client.go
```diff
@@ -379,34 +379,53 @@ func (c *Client) Mail(from string, opts *MailOptions) error {
 	if err := c.hello(); err != nil {
 		return err
 	}
-	cmdStr := "MAIL FROM:<%s>"
+
+	var sb strings.Builder
+	sb.Grow(510 + 14 + 26 + 11 + 9 + 9 + 39 + 500)
```
emersion, line 384: Hm, this seems a bit like micro-optimization with these magic constants... I'd prefer to leave this out or just pick a reasonably high enough power of 2 value.

emersion:
These are mostly just nits, I think this is overall good to go.

smtp.go
ikedas: These follow the naming conventions adopted by @foxcpp in #135. I have no particular preference.

conn.go
ikedas: I'm not a fan too, but could't think of a better way.

smtp.go
emersion: I've went with the prefix everywhere.

emersion (approved):
LGTM, thank you!

PR #237, opened 2023-11-01
 Support pre-SMTP connection rejection

kshangx:
This pull request adds the capability to reject connections to the server before any SMTP communication takes place. The primary motivation behind this feature is to provide control over incoming connections, allowing us to restrict access to our SMTP server to only specific IP ranges (in our use case, the IP ranges will be fetched dynamically from a remote source)

### DIFF
--- a/server.go
+++ b/server.go
@@ -22,7 +22,7 @@ var (
 // A function that creates SASL servers.
 type SaslServerFactory func(conn *Conn) sasl.Server
 
-// A function that
+// A function that checks if a connection should be allowed.
 type AllowConnFunc func(conn *Conn) bool
 
 // Logger interface is used by Server to report unexpected internal errors.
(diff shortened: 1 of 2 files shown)


emersion:
I'd rather not add this: instead, I would recommend you implement your own `net.Listener` which wraps the stdlib's and performs its custom logic on top. This is more flexible and doesn't require the library to add new API.

kshangx:
Thanks @emersion We will look into that option.

kshangx:
Hi @emersion We would like to reopen this PR and see if you have other suggestions.

We tried to implement the `net.Listener` and return error on `Accept` if connection ip is not allowed, but this will cause the server to drop (https://github.com/emersion/go-smtp/blob/master/server.go#L145). If we return a temp net error, there will be temp delay that is not ideal for our use case.

emersion:
Instead, of returning an error, you can continue to wait for the next connection to come in.

kshangx:
Oh got it, thanks

PR #239, opened 2023-11-08
chore: Remove 'Blame it on the weather' error part.

rlejeune74:


emersion:
Why was this closed?

Applied in 49bf39f527fa.

rlejeune74:
This MR wasn't first intend to applied on top of your repo but on our forked one.
(Mistake from me while  creating the MR).
But glad you applied it in your side as well ;)

Issue #243, opened 2023-12-22
[Security] Implement strict CRLF mode


foxcpp:
See https://www.postfix.org/smtp-smuggling.html

We probably should implement the same "Long-term fix".

emersion:
Yeah, seen that… Indeed, we probably want a strict mode at least on-by-default.

PR #244, opened 2023-12-25
Remove DotLF to EOFState case

blmayer:
Hi, this should be enough for closing #243. However, since this is a small code change, I did not test anything.

### DIFF
--- a/data.go
+++ b/data.go
@@ -112,11 +112,6 @@ func (r *dataReader) Read(b []byte) (n int, err error) {
 				r.state = stateDotCR
 				continue
 			}
-			if c == '\n' {
-				r.state = stateEOF
-				continue
-			}
-
 			r.state = stateData
 		case stateDotCR:
 			if c == '\n' {
--- a/server_test.go
+++ b/server_test.go
@@ -733,6 +733,37 @@ func TestServer_tooLongMessage(t *testing.T) {
 	}
 }
 
+// See https://www.postfix.org/smtp-smuggling.html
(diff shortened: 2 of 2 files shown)


emersion:
I've written a small test for this here: https://github.com/emersion/go-smtp/tree/smtp-smuggling-test

Could you cherry-pick that commit and make sure that it passes?

blmayer:
Yes.

blmayer:
The test failed due to a timeout. I think `scanner.Scan()` doesn't return, as it is expecting the correct end. I added a smaller timeout to that test, and removed the check for an incorrect end, since the server doesn't error out: it is correctly expecting more data.

See if you agree with this.

emersion:
Ah right, of course. I've pushed another variant of the test which shouldn't cause a timeout.

blmayer:
Cool, much better, I had to adjust the check's text.

emersion:
For reference, excerpt from the RFC:

> The custom of accepting lines ending only in <LF>, as a concession to
> non-conforming behavior on the part of some UNIX systems, has proven
> to cause more interoperability problems than it solves, and SMTP
> server systems MUST NOT do this, even in the name of improved
> robustness.  In particular, the sequence "<LF>.<LF>" (bare line
> feeds, without carriage returns) MUST NOT be treated as equivalent to
> <CRLF>.<CRLF> as the end of mail data indication.

emersion:
Thanks for the fix!

PR #249, opened 2024-01-20
Fix Conn.Hostname being empty when NewSession is called

foxcpp:
Conn.Hostname is the only way hostname can be read in NewSession. https://github.com/emersion/go-smtp/commit/d62ec8c7d00892343e81b21ad4fb33bcb7a900bb breaks this by setting it _after_ NewSession.


emersion (approved):
Thanks!

PR #250, opened 2024-01-21
Parse DATA\r\n\r\n.\r\n as \r\n\r\n message

foxcpp:
Note that this case is not exactly "no mail data" case referenced in RFC 5321. It is a message containing two empty lines (not one). FWIW This is what go-smtp client send if data writer is closed without writing to it - not sure where extra CRLF comes from.

### DIFF
--- a/data.go
+++ b/data.go
@@ -106,6 +106,10 @@ func (r *dataReader) Read(b []byte) (n int, err error) {
 				r.state = stateDot
 				continue
 			}
+			if c == '\r' {
+				r.state = stateCR
+				break
+			}
 			r.state = stateData
 		case stateDot:
 			if c == '\r' {
--- a/server_test.go
+++ b/server_test.go
@@ -636,6 +636,45 @@ func TestServer_LFDotLF(t *testing.T) {
 	}
 }
 
+func TestServer_EmptyMessage(t *testing.T) {
+	be, s, c, scanner := testServerAuthenticated(t)
+	defer s.Close()
+	defer c.Close()
+
+	io.WriteString(c, "MAIL FROM:<root@nsa.gov>\r\n")
+	scanner.Scan()
+	if !strings.HasPrefix(scanner.Text(), "250 ") {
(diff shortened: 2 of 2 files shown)


data.go
```diff
@@ -106,6 +106,10 @@ func (r *dataReader) Read(b []byte) (n int, err error) {
 				r.state = stateDot
 				continue
 			}
+			if c == '\r' {
+				r.state = stateCR
+				break
```
emersion, line 111: We should probably keep the CR, no? IOW, remove this `break`, just like in the `stateData` case?

emersion: Oh no, I'm confused. Nevermind!

emersion (approved):
Thanks!

j0sh:
Thanks for this fix.

Would be great to get a release with this PR in, because it also fixes a more general issue where the body might have a trailing CRLF, in addition to the actual end sequence. This would make the reader hang.

RFC5321 encourages receivers to be tolerant of trailing whitespace.

Example reproduction (breaks before this PR, fixed after)

```golang
// data.go

func stringDataReader(s string) *dataReader {
  return &dataReader{ r: bufio.NewReader(strings.NewReader(s)) }
}

func main(){

  // original message with trailing crlf
  msg := "to: a@b.com\r\nsubject:test\r\n\r\nbody body\r\n"
  // append crlf-dot-crlf as per smtp
  msg = msg + "\r\n.\r\n"

  out := make([]byte, len(msg))
  dr := stringDataReader(msg)
  n, err := dr.Read(out)
  if err != nil {
    if io.EOF == err {
        fmt.Println("done reading")
        return
    }
    fmt.Println("error reading", err)
    return
  }
  fmt.Printf("not done reading - read %d out of %d - %#v\n", n, len(msg), dr)
}
```

emersion:
Indeed!

I've just shipped a bugfix release.

nerzhul:
Thanks, i hope this fix the read timeout i have on some mails, let's test it :)

Issue #170, opened 2021-11-29
server: Impossible to disable PLAIN auth


kayrus:
PLAIN auth is enabled by default and it is not possible to disable it:

https://github.com/emersion/go-smtp/blob/30169acc42e795e5d35ce901c8387950b103dfd9/server.go#L86-L87

it also doesn't look correct that `Session` interface now requires the `AuthPlain` method (#146). Why doesn't it require other auth methods? Why don't rely on sasl auth methods defined explicitly?

emersion:
go-imap v2's design: https://pkg.go.dev/github.com/emersion/go-imap/v2/imapserver#SessionSASL

emersion:
https://github.com/emersion/go-smtp/pull/254 fixes this.

PR #251, opened 2024-02-21
server: replace EnableAuth with AuthSession

emersion:
Closes: https://github.com/emersion/go-smtp/issues/170

Future work:

- Drop `Session.AuthPlain`
- Drop `Server.AuthDisabled`

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -2,6 +2,8 @@ package smtp
 
 import (
 	"io"
+
+	"github.com/emersion/go-sasl"
 )
 
 var (
@@ -19,6 +21,11 @@ var (
 		Code:         502,
 		EnhancedCode: EnhancedCode{5, 7, 0},
 		Message:      "Authentication not supported",
+	}
+	ErrAuthUnknownMechanism = &SMTPError{
+		Code:         504,
+		EnhancedCode: EnhancedCode{5, 7, 4},
+		Message:      "Unsupported authentication mechanism",
 	}
 )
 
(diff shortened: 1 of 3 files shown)


yusufozturk:
@emersion is there any documentation or test update to see how we can upgrade from older versions to this version?

image

Our build job was broken after package upgrades.

yusufozturk:
Just feedback:

Maybe consider keeping old API methods for a while but display a deprecated warning in the console. Link the old methods to the new ones so that the old methods serve only as a compatibility bridge. This way, users can see how they should upgrade their code.

Thanks for this great library and your efforts to keep it up-to-date. @emersion

mgnsk:
I recently went through the upgrade. Basically, the server no longer cares about authentication, it is up to sessions to implement either `Session` or `AuthSession`.

The server checks if the session implements `AuthSession` and `AuthMechanisms()` returns a non-empty slice and advertises the `AUTH` capability if so. Similarly, if the session implements `AuthSession`, it calls `AuthSession.Auth()` during the `AUTH` command.

To avoid implementing two different sessions, we only implemented a single session that always implements `AuthSession` and manually returns `ErrAuthUnknownMechanism` from its `Auth()` method if no mechanisms are enabled.

yusufozturk:
@mgnsk Thanks!

Can you share some code blocks how do you implement a single session? For the session, I think I need to change Backend implementation but I don't know how I can do that. Thanks!

emersion:
The release notes have a high-level overview of the breaking changes: https://github.com/emersion/go-smtp/releases/tag/v0.21.0

I considered leaving the old API as deprecated but it wasn't possible to do so properly without a lot of churn. Keep in mind that I'm maintaining this library during my free time as a volunteer.

PR #253, opened 2024-02-29
feat: adding the smtp auth login mechanism

buroa:
This is needed so I can finish writing unit tests for https://github.com/foxcpp/maddy/pull/683

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -37,7 +37,7 @@ type Session interface {
 	// Free all resources associated with session.
 	Logout() error
 
-	// Authenticate the user using SASL PLAIN.
+	// Authenticate the user using SASL LOGIN/PLAIN mechanism.
 	AuthPlain(username, password string) error
 
 	// Set return path for currently processed message.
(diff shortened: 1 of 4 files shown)


emersion:
Sorry, but I don't want to support the legacy LOGIN mechanism by default. Library users to register their own SASL mechanisms if they want to.

buroa:
> Sorry, but I don't want to support the legacy LOGIN mechanism by default. Library users to register their own SASL mechanisms if they want to.

Are you fucking serious?

buroa:
@emersion You need to get off of Github if you are not going to be open to conversations regarding open-source software. Closing issues within seconds of it being open is toxic. I'll make sure people know this when looking at your products in the future.

emersion:
I don't believe closing pull requests for undesirable features is being toxic. I generally do leave pull requests opened when I am not 100% sure that the feature is undesirable, but this is not the case here, the topic at hand has been discussed before in the context of go-sasl and is well-understood.

Being an open-source maintainer doesn't mean that I _have to_ accept all features. I won't comment on the aggressive threat, that's enough for me, I do not wish to interact further with someone with your mindset.

PR #255, opened 2024-03-23
Server: The value of BODY parameter is case-insensitive

ikedas:
Currently,
```code
MAIL FROM:<alice@wonderland.book> BODY=8bitMIME
```
is responded by `500 5.5.4 Unknown BODY value`.

This PR fixes it to make the parameter value case-insensitive.

___
In this occasion, it also makes status code for invalid parameter value `501` as RFC describes.


emersion (approved):
LGTM, thanks!

PR #257, opened 2024-04-03
Update example server auth for AuthSession interface in 0.21.0 update

jasonfriedland:
Update for docs at: https://pkg.go.dev/github.com/emersion/go-smtp#example-Server


emersion (approved):
Thanks!

PR #259, opened 2024-04-11
Fix anonymous authentication with empty trace information string

### LINKED ISSUE #258: Client anonymous authentication not working when sending empty trace string
Example code:
```
client.Auth(sasl.NewAnonymousClient(""))
```

Output from a test server:
```
220 localhost ESMTP Service Ready
EHLO localhost
250-Hello localhost
250-PIPELINING
250-8BITMIME
250-ENHANCEDSTATUSCODES
250-CHUNKING
250-AUTH ANONYMOUS
250-SIZE 1024
250 LIMITS RCPTMAX=10
AUTH ANONYMOUS
334
*
501 5.0.0 Negotiation cancelled
smtp/server 2024/04/10 19:25:09 handler error: read tcp 127.0.0.1:10025->127.0.0.1:58986: wsarecv: An existing connection was forcibly closed by the remote host.
```

Output from a test client:

```
220 localhost ESMTP Service Ready
EHLO localhost
250-Hello localhost
250-PIPELINING
250-8BITMIME
250-ENHANCEDSTATUSCODES
250-CHUNKING
250-AUTH ANONYMOUS
250-SIZE 1024
250 LIMITS RCPTMAX=10
AUTH ANONYMOUS
334
*
501 5.0.0 Negotiation cancelled
2024/04/10 19:25:09 sasl: unexpected server challenge
```

albertony:
With anonymous authentication according to RFC4505 the trace information string is optional, and SMTP authentication extension described in RFC4954 states that:

    If the client is transmitting an initial response of zero
    length, it MUST instead transmit the response as a single
    equals sign ("=").  This indicates that the response is
    present, but contains no data.

Fixes #258

### DIFF
--- a/client.go
+++ b/client.go
@@ -356,8 +356,13 @@ func (c *Client) Auth(a sasl.Client) error {
 	if err != nil {
 		return err
 	}
-	resp64 := make([]byte, encoding.EncodedLen(len(resp)))
-	encoding.Encode(resp64, resp)
+	var resp64 []byte
+	if len(resp) > 0 {
+		resp64 = make([]byte, encoding.EncodedLen(len(resp)))
+		encoding.Encode(resp64, resp)
+	} else if resp != nil {
+		resp64 = []byte{'='}
+	}
 	code, msg64, err := c.cmd(0, strings.TrimSpace(fmt.Sprintf("AUTH %s %s", mech, resp64)))
 	for err == nil {
 		var msg []byte
--- a/conn.go
+++ b/conn.go
@@ -773,11 +773,15 @@ func (c *Conn) handleAuth(arg string) {
 	// Parse client initial response if there is one
 	var ir []byte
 	if len(parts) > 1 {
-		var err error
-		ir, err = base64.StdEncoding.DecodeString(parts[1])
-		if err != nil {
-			c.writeResponse(454, EnhancedCode{4, 7, 0}, "Invalid base64 data")
-			return
+		if parts[1] == "=" {
+			ir = []byte{}
+		} else {
+			var err error
+			ir, err = base64.StdEncoding.DecodeString(parts[1])
+			if err != nil {
+				c.writeResponse(454, EnhancedCode{4, 7, 0}, "Invalid base64 data")
+				return
+			}
 		}
 	}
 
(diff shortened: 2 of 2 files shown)


emersion:
These changes look good to me overall, just one comment: I think we probably need to also handle `=` when sent as a regular response (not an initial response)? See the other `base64.StdEncoding.DecodeString` call done below in the same function, inside the `for` loop.

albertony:
I agree. Force pushed an update. I tested it with openssl s_client against my go-smtp based test server, and I assume a sequence like this is what you were thinking of:

```
EHLO localhost
250-Hello localhost
250-PIPELINING
250-8BITMIME
250-ENHANCEDSTATUSCODES
250-CHUNKING
250-AUTH ANONYMOUS
250-SIZE 1024
250 LIMITS RCPTMAX=10
AUTH ANONYMOUS
334
=
```

PTAL

client.go
```diff
@@ -356,8 +356,13 @@ func (c *Client) Auth(a sasl.Client) error {
 	if err != nil {
 		return err
 	}
-	resp64 := make([]byte, encoding.EncodedLen(len(resp)))
-	encoding.Encode(resp64, resp)
+	var resp64 []byte
+	if len(resp) > 0 {
+		resp64 = make([]byte, encoding.EncodedLen(len(resp)))
+		encoding.Encode(resp64, resp)
+	} else {
```
emersion, line 363: Oh, one last thing: the go-sasl API has the same nil vs empty slice footgun in the `Client` interface. Namely, here `resp = nil` means no initial response, while `resp = []byte{}` means an empty initial response.

The fix would be to replace this line with `} else if resp != nil {` I believe.

albertony: Hm, ok. Force pushed update. I haven't tested it though, wasn't immediately sure how to (and no time to investigate more right now).

emersion: That's OK. Would need to find a SASL mechanism with no initial response, I don't know of any.

It seems like the push hasn't gone through?

albertony: Ah, sorry. Forgot the commit in between add and push... 🥵

albertony: Was that change what you intended?

emersion (approved):
LGTM, thank you!

emersion:
Oh, CI is unhappy now… Maybe the test actually tests the no initial response case, as opposed to the empty initial response one?

emersion:
Seems like it:

```go
func (toServerEmptyAuth) Start() (proto string, toServer []byte, err error) {
	return "FOOAUTH", nil, nil
}
```

emersion:
I've removed the test changes, and fixed up the outdated comments in 029f7513c5fe.

albertony:
Ahhh, real sorry about that! I've been running that many times, and adjusted it in a previous iteration, but probably forgot about it during these last iterations. Thanks for fixing it (I'm not at computer right now anyway).

emersion:
Yeah, no worries whatsoever, it's been a ride untangling this mess!

PR #260, opened 2024-04-17
Do not use HELLO cmd as fallback of EHLO when server responds with code 421

diogomr:
Change inspired by https://github.com/PHPMailer/PHPMailer/issues/2189

RFC1869, Section 4.5 (https://datatracker.ietf.org/doc/html/rfc1869#section-4.5) states that the server will return the code 421 if the SMTP server is no longer available

This change fixes an issue where the actual error response from a failed EHLO was not surfaced due to always being overridden by the HELLO response.

A real case scenario of this issue:
```
220 smtp.hostedemail.com ESMTP
EHLO goodvouch.com
421 Service not available, closing transmission channel
HELO goodvouch.com
```

The error surfaced by `client.Hello` is `EOF` instead of `421 Service not available, closing transmission channel`


diogomr:
Had a bit of an hard time figuring how to best test this change, so please let me know if you prefer a different style.

Also feel free to edit anything directly if you don't like the style.

emersion:
Hm, maybe we can tighten this even further and only attempt HELO if the server returns a 500 error ("command not recognized")?

diogomr:
> Hm, maybe we can tighten this even further and only attempt HELO if the server returns a 500 error ("command not recognized")? See https://datatracker.ietf.org/doc/html/rfc5321#section-3.2
> 
> Or do you think this won't work in some cases?

Tbh my knowledge of SMTP is not very deep, but that sounds ok.

The test suite on `client_test.go` uses a `502` (`502 EH?`) to test the fallback to `HELO`. Do you have context on why the 502 was used?

emersion:
Hm, reading [section 4.2.4](https://datatracker.ietf.org/doc/html/rfc5321#section-4.2.4), seems like 502 means that the server knows that the command is defined in an RFC but it's not implemented. I suppose it should be fine to catch that one as well.

client_test.go
```diff
@@ -509,6 +509,38 @@ var helloClient = []string{
...
+	defer c.Close()
+	c.serverName = "fake.host"
+	c.localName = "customhost"
+
+	err := c.Hello("customhost")
+	if err == nil {
+		t.Errorf("Expected Hello to fail")
+	}
+
+	if err.Error() != "SMTP error 421: Service not available, closing transmission channel" {
```
emersion, line 533: Can we check for an `SMTPError` with `errors.As` here? To make the check more robust.

client_test.go
```diff
@@ -242,7 +242,7 @@ func TestBasic_SMTPError(t *testing.T) {
 func TestClient_TooLongLine(t *testing.T) {
 	faultyServer := []string{
 		"220 mx.google.com at your service\r\n",
-		"220 mx.google.com at your service\r\n",
+		"250 mx.google.com at your service\r\n",
```
emersion, line 245: Is this change intentional?

diogomr: Yes, this test started failing as a side effect of the change, but I think the test setup was wrong.

The first `220 mx.google.com at your service\r\n` would be consumed by `greet`, the second `220 mx.google.com at your service\r\n` would be consumed by `ehlo` which interprets anything other than 250 as an error and would fallback to using `helo`, which would capture the long line. 

With the changes I introduced in this PR, the `220` code won't trigger the `helo` and the long line is never consumed.

Does that make sense? My understanding is that the change is only breaking test assumptions.

client_test.go
diogomr: Added the check on https://github.com/emersion/go-smtp/pull/260/commits/0ebd8acee04a66ca8325fbb94895e329d49c4703

diogomr: Added the check on https://github.com/emersion/go-smtp/pull/260/commits/e1cdaa56f2a4b3da47b689426a66ab053df6b7f1

client_test.go
emersion: Ah yes indeed, 220 is not a valid response to EHLO or HELO. Good catch! The last line also has the same issue.

The message after the SMTP code is still confusing though: it looks like a greeting when it's an OK response.

I've fixed these issues in 8fc2197231c8.

emersion (approved):
Thanks!

PR #262, opened 2024-04-23
Add remote addr in error handler log

sapmli:
When an incoming connection is not working, e.g. due to a missing proxy protocol header, there is currently no chance to catch details on server side, e.g. to debug where it came from.

This PR adds the remote address to the log line, making it more consistent with other logs like `panic serving %v: %s...`

Another solution might be to pass the connection as first parameter to the logger interface functions, but it's a heavier change and it might not work for situations where the connection could not be established at all.


emersion (approved):
Thanks!

PR #266, opened 2024-07-04
client: add new functions to specify options

rjarry:
Expose a new set of functions to create a client with custom options.


rjarry:
Hi @emersion, do you have any remarks?

Since the 0.21 release, it is not possible to connect with STARTTLS and a custom local name other than `localhost`. Some MTAs enforce clients to expose their real hostname and reject email when the `HELO` contains a non FQDN.

```
send:: conn.Rcpt: SMTP error 504: <localhost>: Helo command rejected: need fully qualified hostname
```

If this PR could be applied and included in a tagged release it would be awesome :)

Thanks!

emersion:
This commit should already fix your issue: https://github.com/emersion/go-smtp/commit/75e52af41ff5fd7fd8e149742b5476a2a47b564f

This PR adds a more convenient way to set client options but doesn't add any new functionality - I'm not 100% sure it's a good idea to have 2 ways of doing the same thing...

rjarry:
Ok, this should do fine. However https://github.com/emersion/go-smtp/commit/75e52af41ff5fd7fd8e149742b5476a2a47b564f it is still on master, not included in any tag yet. Could you cut a `0.21.3` version including this?

Thanks!

emersion:
This was more tricky than just releasing a new bugfix release, because master also has another commit with breaking changes. I've established a branch for v0.21.x and released 0.21.3.

emersion:
Yeah, please keep it open.

If I had to design an API from scratch, I'd probably only expose options and not expose public fields in `Client`.

lukeliu-cybavo:
the default timeout toooooo long, we need this PR or something like this

JokerCatz:
the default timeout toooooo long, we need this PR or something like this

PR #268, opened 2024-08-24
XOORG Server Support

ml1nk:
These changes allows the smtp server to advertise XOORG capability and extract it into the options. There is a simple test case added and it is tested against Exchange Online with a Outgoing Connector.

Some information about XOORG and it's usages:
https://techcommunity.microsoft.com/t5/exchange-team-blog/advanced-office-365-routing-locking-down-exchange-on-premises/ba-p/609238

I will also add XOORG client support next week, to be able to use it in combination with a Exchange Online Inbound Connector.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -276,6 +276,9 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 	if c.server.EnableDSN {
 		caps = append(caps, "DSN")
 	}
+	if c.server.EnableXOORG {
+		caps = append(caps, "XOORG")
+	}
 	if c.server.MaxMessageBytes > 0 {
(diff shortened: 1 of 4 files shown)


emersion:
Sorry, I'd rather not maintain some Microsoft-specific proprietary features.

ml1nk:
I understand your reasoning behind it, but I would argue that
- the changes in the server (and client later on) are very small without anything special, so there is no big maintenance to be done
- it's a non-standard but well documented extension which is very helpful in O365/Hybrid Exchange environments whenever you need to work with connectors
- it's hidden behind a flag, so no harm is done if not used

Issue #269, opened 2024-08-29
EHLO after RSET


sapmli:
Currently, go-smtp starts a new session for every EHLO command it receives.
So, any user information, e.g. from AUTH LOGIN, gets lost.

[RFC 5321](https://datatracker.ietf.org/doc/html/rfc5321#section-4.1.4) states:
> An EHLO command MAY be issued by a client later in the session.  If
   it is issued after the session begins and the EHLO command is
   acceptable to the SMTP server, the SMTP server MUST clear all buffers
   and reset the state exactly as if a RSET command had been issued.  In
   other words, the sequence of RSET followed immediately by EHLO is
   redundant, but not harmful other than in the performance cost of
   executing unnecessary commands.

Probably this means the new EHLO argument should be handed over to an existing session, instead of starting a new session. Clients may issue an EHLO after RSET, to set a different client information for the next mail in the same connection.

It's currently also not possible to send a new EHLO after RSET in outgoing connections using the [SMTP client](https://github.com/emersion/go-smtp/blob/master/client.go#L204-L212).

emersion:
Indeed, seems like `Conn.reset()` should be used instead of creating a new session if we already have one.

PR #270, opened 2024-09-02
Allow custom EHLO after RSET again

sapmli:
Resolves [EHLO after RSET #269](https://github.com/emersion/go-smtp/issues/269)

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -229,14 +229,21 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 	// NewSession can access it via Conn.Hostname.
 	c.helo = domain
 
-	sess, err := c.server.Backend.NewSession(c)
-	if err != nil {
-		c.helo = ""
-		c.writeError(451, EnhancedCode{4, 0, 0}, err)
-		return
-	}
+	// RFC 5321: "An EHLO command MAY be issued by a client later in the session"
+	if c.session != nil {
+		// RFC 5321: "... the SMTP server MUST clear all buffers
+		// and reset the state exactly as if a RSET command has been issued."
+		c.reset()
+	} else {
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -229,14 +229,25 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
...
-	if err != nil {
-		c.helo = ""
-		c.writeError(451, EnhancedCode{4, 0, 0}, err)
-		return
-	}
+	// RFC 5321: "An EHLO command MAY be issued by a client later in the session"
+	if c.session != nil {
+		// RFC 5321: "If it is issued after the session begins
+		// and the EHLO command is acceptable to the SMTP server, ..."
+		if !c.fromReceived {
```
emersion, line 236: Hm, I don't quite get why this condition is here. Why do we need to skip `reset()` when the client has sent `MAIL FROM`?

sapmli: It's what I found in the session that may be usable for the condition "if it is issued after the session begins"

emersion: ["Session begins"](https://datatracker.ietf.org/doc/html/rfc5321#section-3.1) just means that a TCP connection is opened no?

sapmli: Intersting question. If so, then do you think that an EHLO after RSET should not be passed to the session as Reset()?

emersion: My point is that we should probably always call `c.reset()`, even in the case where `MAIL FROM` was sent by the client (that's what `fromReceived` means).

sapmli: Okay, I can do that. I'll update the PR.

sapmli: But feel free to adapt it to your likings and code style etc.

emersion (approved):
Thanks!

I've re-organized the changes into two commits.

PR #272, opened 2024-09-16
Add checking for auth to Server example.

alext:
This was originally added to the `README.md` in #222, but was lost when the examples were removed from there.

Re-add it so that it's clearer that this library doesn't enforce authentication by default.


emersion (approved):
Thanks!

PR #274, opened 2024-11-02
Add a status hook

lyda:
Run a function when sending an SMTP status code.  Use for logging, metrics, debugging, etc.

I'm specifically doing this because I want to export out counts of SMTP server status and extended status codes as prometheus metrics. Folks could use it for other things as well - logging, other metrics systems, etc.


emersion:
Sorry, I don't want to maintain this feature.

Issue #275, opened 2024-11-02
Incorrect handling of multiline errors


pch:
I believe that multiline errors aren't being handled properly. Consider this example:

```go
ErrRecipientDomainMxCheckFailed = &smtp.SMTPError{
	Code:         550,
	EnhancedCode: smtp.EnhancedCode{5, 1, 2},
	Message:      "Recipient domain MX check failed\nthis is another line\nand another one",
}
```

Returning this error will cause the following exchange:

```
 ~> RCPT TO:<someone@invalidemaildomain.com>
<~* 550 5.1.2 Recipient domain MX check failed
 ~> QUIT
<~  this is another line
<~  and another one
<~  221 2.0.0 Bye
```

This causes incorrect/premature QUIT and breaks client response readers.

Correct response should be:

```
<~* 550-5.1.2 Recipient domain MX check failed
<~* 550-5.1.2 this is another line
<~* 550 5.1.2 and another one
```

The issue seems to be here:

```go
func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
```

This function looks to be handling multiline responses correclty, but it's expecting `text` to be multiple arguments and it's getting a single multiline string.

A (naive) fix would be something like:

```diff
-func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
+func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text string) {
+       lines := strings.Split(text, "\n")
+
        // TODO: error handling
        if c.server.WriteTimeout != 0 {
                c.conn.SetWriteDeadline(time.Now().Add(c.server.WriteTimeout))
@@ -1236,13 +1238,13 @@ func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
                }
        }

-       for i := 0; i < len(text)-1; i++ {
-               c.text.PrintfLine("%d-%v", code, text[i])
+       for i := 0; i < len(lines)-1; i++ {
+               c.text.PrintfLine("%d-%v", code, lines[i])
        }
        if enhCode == NoEnhancedCode {
-               c.text.PrintfLine("%d %v", code, text[len(text)-1])
+               c.text.PrintfLine("%d %v", code, lines[len(text)-1])
        } else {
-               c.text.PrintfLine("%d %v.%v.%v %v", code, enhCode[0], enhCode[1], enhCode[2], text[len(text)-1])
[...]

PR #279, opened 2025-03-29
server: handle newline characters in error messages

qugu2427:
This very simple pr allows multi line error messages delimited with \n, per Issue #275.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1236,13 +1236,17 @@ func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
 		}
 	}
 
-	for i := 0; i < len(text)-1; i++ {
+	// transform each single line with \n, into seperate lines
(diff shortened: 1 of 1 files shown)


conn.go
```diff
@@ -1236,13 +1236,17 @@ func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
 		}
 	}
 
-	for i := 0; i < len(text)-1; i++ {
+	// transform each single line with \n, into separate lines
+	text = strings.Split(strings.Join(text, "\n"), "\n")
+
+	textLen := len(text) - 1
```
emersion, line 1242: This isn't the length, since there is this "-1". Maybe should be renamed to `lastLineIndex`?

qugu2427: works for me

emersion (approved):
Thanks!

PR #280, opened 2025-04-05
Implement RRVS

qugu2427:
This pr implements [RRVS](https://www.rfc-editor.org/rfc/rfc7293.html) for both the client and server.

### DIFF
--- a/client.go
+++ b/client.go
@@ -81,7 +81,7 @@ func DialTLS(addr string, tlsConfig *tls.Config) (*Client, error) {
 	return client, nil
 }
 
-// DialStartTLS retruns a new Client connected to an SMTP server via STARTTLS
+// DialStartTLS returns a new Client connected to an SMTP server via STARTTLS
 // at addr. The addr must include a port, as in "mail.example.com:smtp".
 //
 // A nil tlsConfig is equivalent to a zero tls.Config.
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -530,6 +530,11 @@ func (c *Client) Rcpt(to string, opts *RcptOptions) error {
 			fmt.Fprintf(&sb, " ORCPT=%s;%s", string(opts.OriginalRecipientType), enc)
 		}
 	}
+	if _, ok := c.ext["RRVS"]; ok && opts != nil {
+		if !opts.RequireRecipientValidSince.Equal(time.Time{}) {
```
emersion, line 534: `IsZero()` is a simpler way to write `.Equal(time.Time{})`.

emersion, line 534: This condition can be combined with the `if` right above.

conn.go
```diff
@@ -716,6 +719,20 @@ func (c *Conn) handleRcpt(arg string) {
...
 			opts.OriginalRecipientType = aType
 			opts.OriginalRecipient = aAddr
+		case "RRVS":
+			if !c.server.EnableRRVS {
+				c.writeResponse(504, EnhancedCode{5, 5, 4}, "RRVS is not implemented")
+				return
+			}
+			if dateTimeEnd := strings.Index(value, ";"); dateTimeEnd != -1 {
+				value = value[:strings.Index(value, ";")]
+			}
```
emersion, line 729: A simpler way to write this:

```go
value, _ = strings.Cut(value, ";") // discard the no-support action
```

smtp.go
```diff
@@ -91,4 +93,8 @@ type RcptOptions struct {
 	// Original recipient set by client.
 	OriginalRecipientType DSNAddressType
 	OriginalRecipient     string
+
+	// Time value of the RRVS= argument
+	// default time.Time value (Jan 1 1970) if not set by client
```
emersion, line 98: "default time.Time value (Jan 1 1970) if not set by client" is confusing: the zero time is not in year 1970, it's in year 1.

I'd suggest replacing the sentence with "Left as the zero time if unset." (or dropping the sentence completely).

emersion (approved):
Thanks!

PR #283, opened 2025-06-21
Implement DELIVERBY (rfc2852)

qugu2427:


smtp.go
```diff
@@ -84,6 +88,19 @@ const (
...
 
+type DeliverByMode string
+
+const (
+	DeliverByNotify DeliverByMode = "N"
+	DeliverByReturn DeliverByMode = "R"
+)
+
+type DeliverByOptions struct {
+	ByTime  int64
```
emersion, line 99: Can we make this a `time.Duration`?

smtp.go
```diff
@@ -84,6 +88,19 @@ const (
...
+	ByMode  DeliverByMode
+	ByTrace bool
```
emersion, line 101: Nit: I think we can drop the `By` prefix for these fields, since it's already in the struct name?

server.go
```diff
@@ -61,6 +61,16 @@ type Server struct {
...
 
+	// Advertise DELIVERBY (RFC 2852) capability.
+	// Should be used only if backend supports it.
+	EnableDELIVERBY bool
+
+	// The minimum time, in seconds, that a client
+	// may specify in the BY argument.
+	// A zero value indicates no set minimum.
+	// Only use if DELIVERBY is enabled.
+	MinimumDeliverByTime int64
```
emersion, line 72: Ditto: can we make this a `time.Duration`?

conn.go
```diff
@@ -731,6 +738,21 @@ func (c *Conn) handleRcpt(arg string) {
...
+			if !c.server.EnableDELIVERBY {
+				c.writeResponse(504, EnhancedCode{5, 5, 4}, "DELIVERBY is not implemented")
+				return
+			}
+			deliverBy := parseDeliverByArgument(value)
+			if deliverBy == nil {
+				c.writeResponse(501, EnhancedCode{5, 5, 4}, "Malformed BY parameter value")
+				return
+			}
+			if c.server.MinimumDeliverByTime != 0 && deliverBy.ByTime < c.server.MinimumDeliverByTime {
```
emersion, line 751: Should we check for R mode here?

> Any attempt by a client to specify a by-mode of "R" and a by-time strictly less than this limit, min-by-time, will be rejected with a permanent failure (55z) reply code.

parse.go
```diff
@@ -74,6 +75,32 @@ func parseHelloArgument(arg string) (string, error) {
...
+// Returns pointer to options or nil if invalid.
+func parseDeliverByArgument(arg string) *DeliverByOptions {
+	secondsStr, modeStr, ok := strings.Cut(arg, ";")
+	if !ok {
+		return nil
+	}
+	traceValue := strings.HasSuffix(modeStr, "T")
+	if traceValue {
+		modeStr = strings.TrimSuffix(modeStr, "T")
+	}
```
emersion, line 88: Nit: this can be simplified with `strings.CutSuffix`.

emersion (commented):
Thanks for the patch! Here are a few comments, but overall this looks pretty good!

client_test.go
```diff
@@ -1079,3 +1079,39 @@ func TestClientRRVS(t *testing.T) {
...
+	}{
+		strings.NewReader(server),
+		&wrote,
+	}
+	c := NewClient(fake)
+	c.didHello = true
+	c.ext = map[string]string{"DELIVERBY": ""}
+	c.Rcpt("root@nsa.gov", &RcptOptions{
+		DeliverBy: &DeliverByOptions{
+			Time:  time.Duration(100) * time.Second,
```
emersion, line 1108: Nit: no need for `time.Duration` here, this should work:

```go
Time: 100 * time.Second
```

emersion (approved):
Thank you!

PR #284, opened 2025-07-05
Implement MT-PRIORITY (RFC 6710)

qugu2427:


smtp.go
```diff
@@ -101,6 +102,15 @@ type DeliverByOptions struct {
 	Trace bool
 }
 
+type PriorityProfile string
+
+const (
+	NONE       PriorityProfile = ""
+	MIXER      PriorityProfile = "MIXER"
+	STANAG4406 PriorityProfile = "STANAG4406"
+	NSEP       PriorityProfile = "NSEP"
```
emersion, line 111: Naming:

- Can we prefix the priority profile names with `Priority`? `NONE` alone is too generic.
- Idiomatic Go names don't use UPPERCASE, they use camelCase.

smtp.go
```diff
@@ -117,4 +127,7 @@ type RcptOptions struct {
 
 	// Value of BY= argument or nil if unset.
 	DeliverBy *DeliverByOptions
+
+	// Value of MT-PRIORITY= or nil if unset.
+	MtPriority *int
```
emersion, line 132: `MTPriority` would be a more idiomatic Go name, because the case in acronyms is retained.

emersion (commented):
Apart from the naming comments, this looks good!

PR #285, opened 2025-07-06
Fix noisy error logging for zero-byte connections

mariuso:
Health checks connect and immediately disconnect without sending any data, this creates noisy logs.

This PR fixes it by only logging errors for connections that actually tried to send SMTP commands. Health checks that send nothing are now silent.

**What changed:**
- Track whether a connection sent any data 
- Skip error logging for zero-byte connections
- Real SMTP errors still get logged normally 

Includes a test that proves health checks no longer spam the logs.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -41,9 +41,10 @@ type Conn struct {
 	dataResult      chan error
 	bytesReceived   int64 // counts total size of chunks when BDAT is used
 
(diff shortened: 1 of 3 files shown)


emersion:
Sending nothing should not cause a connection error, it should cause an immediate `io.EOF`.

PR #286, opened 2025-07-06
Add ConnState hook for connection lifecycle monitoring

mariuso:
This PR adds connection lifecycle hooks to enable monitoring, metrics, and debugging of SMTP connections.

**What's new:**
- `ConnState` callback field on Server that gets called when connections change state
- Connection states: New, Active, Auth, Data, StartTLS, Reset, Idle, Error, Closed
- Each state transition triggers the callback with the connection and new state

**Use cases:**
- Track active connections count
- Monitor connection duration and state transitions
- Debug authentication issues and connection problems
- Build metrics dashboards for SMTP server health

**Example:**
```go
server.ConnState = func(conn net.Conn, state smtp.ConnState) {
    switch state {
    case smtp.StateNew:
        activeConnections.Inc()
    case smtp.StateClosed:
        activeConnections.Dec()
    }
}
```

No breaking changes - the callback is optional and everything works exactly the same if you don't set it.

Includes comprehensive tests showing all state transitions.

### DIFF
--- a/server.go
+++ b/server.go
@@ -20,6 +20,44 @@ type Logger interface {
 	Println(v ...interface{})
 }
 
+// ConnState represents the state of a client connection to a server.
+type ConnState int
+
+const (
+	// StateNew represents a new connection that is expected to
+	// send SMTP commands immediately.
+	StateNew ConnState = iota
+	
+	// StateActive represents a connection that has sent the initial
+	// greeting and is actively processing SMTP commands.
+	StateActive
+	
+	// StateAuth represents a connection during SASL authentication handshake.
+	StateAuth
+	
+	// StateData represents a connection receiving message data (DATA command).
+	StateData
+	
+	// StateStartTLS represents a connection during TLS handshake.
+	StateStartTLS
+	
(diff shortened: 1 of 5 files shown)


arp242:
I think `c.setState(StateError)` can be in the `Conn.WriteResponse()`? That is:

	func (c *Conn) writeResponse(code int, enhCode EnhancedCode, text ...string) {
		if code >= 400 {
			c.setState(StateError)
		}

		[..]
	}

There's a bunch of error replies where you're not calling setState(stateError) though, I'm not sure if that's intentional?

server.go
```diff
@@ -131,8 +173,15 @@ func (s *Server) Serve(l net.Listener) error {
...
-			if err != nil {
+			conn := newConn(c, s)
+			
+			// Notify new connection state
+			if s.ConnState != nil {
+				s.ConnState(c, StateNew)
+			}
+			
+			err := s.handleConn(conn)
+			if err != nil && conn.hasReceivedData {
```
arp242, line 184: Doesn't compile because `conn.hasReceivedData` doesn't exist; seems to be a left-over of your other PR (#285)?

PR #287, opened 2025-07-08
Remove superfluous validateLine() calls in sendMail()

arp242:
Client.SendMail() already calls validateLine() via Client.Mail() and Client.Rcpt(), so there is no need to have these checks in sendMail().

Having these checks don't hurt as such, but looking at just the sendMail() function it's not obvious whether these checks are needed if you want to implement your own SendMail()-type function. With this change, you can copy/paste sendMail(), add the smtp. package selector, and modify as needed.

Also removes a superfluous nil check as a bonus.


emersion (approved):
Thanks!

PR #292, opened 2025-08-28
allow custom serverName in client

sapmli:
In order to use a custom dialer with a different timeout for the SMTP client,
we need to be able set the client's serverName after calling `NewClient(conn)` or `NewClientStartTLS(conn, tlsConfig)`,
so that TLS server certificates can still be matched to the hostname.

Other solutions might be to
- Extend the `NewClient...()` calls by an `serverName string` argument (not backwards-compatible)
- Add a `func (c *Client) SetServerName(string)` method

I chose to make the `serverName` field public to minimize changes, be backwards-compatible and consistent with customizing other fields like the `CommandTimeout`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,7 +26,6 @@ type Client struct {
 	// connection later
 	conn       net.Conn
 	text       *textproto.Conn
-	serverName string
 	lmtp       bool
 	ext        map[string]string // supported extensions
 	localName  string            // the name to use in HELO/EHLO/LHLO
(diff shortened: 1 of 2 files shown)


sapmli:
I just noticed that this doesn't work for StartTLS since the`NewClientStartTLS()` already needs the serverName.

emersion:
Can't users already set a custom name by passing a `tls.Config` with an already-populated `ServerName`?

sapmli:
Yes, probably we can do that this way.
I played around with it and it seems to work.

PR #293, opened 2025-09-14
feat: add check command sequence

sashamelentyev:


emersion:
What is the motivation for this?

sashamelentyev:
> What is the motivation for this?

Just implement 503 error code of this https://support.google.com/a/answer/3221692?hl=en

conn.go
```diff
@@ -111,6 +115,13 @@ func (c *Conn) handle(cmd string, arg string) {
 		// These commands are not implemented in any state
 		c.writeResponse(502, EnhancedCode{5, 5, 1}, fmt.Sprintf("%v command not implemented", cmd))
 	case "HELO", "EHLO", "LHLO":
+		if c.commandSequence != 1 {
```
emersion, line 118: I don't think we should enforce an order for HELO/EHLO.

conn.go
```diff
@@ -44,6 +44,8 @@ type Conn struct {
 	fromReceived bool
 	recipients   []string
 	didAuth      bool
+
+	commandSequence int
```
emersion, line 48: Rather than an int with magic values, I would prefer a separate type.

We already have `fromReceived` which covers some of this.

conn.go
```diff
@@ -123,8 +134,22 @@ func (c *Conn) handle(cmd string, arg string) {
 		}
 		c.handleGreet(enhanced, arg)
 	case "MAIL":
+		if c.commandSequence != 2 {
```
emersion, line 137: This logic should be moved to `handle` methods.

emersion (commented):
GMail docs do not define SMTP. The order of commands is defined in https://datatracker.ietf.org/doc/html/rfc5321#section-4.1.4

We do not use conventional commit messages.

sashamelentyev:
@emersion thank you for review!
I close this MR. If need it case may be implement in session

PR #294, opened 2025-09-20
Implemented dynamic domain support

endreszabo:
## What is this

Implemented dynamic domain support that also doubles as a health check endpoint for loadbalancers.
Details:

### Dynamic domain support
Traditionally `conn.Domain` contains the domain that should be the server identifying itself to the connecting clients with during the initial greeting. My PR enabled this field to be dynamic: a function can be specified to generate the value for this field instead of passing the `conn.Domain` value. It is backward compatible as the greeter will send the value of the `conn.Domain` field if the dynamic getter function is not set. This feature is useful if a given server handles traffic for more than one endpoint.

### Health check support
Greeter function will reject the connection with 554 if the dynamic function returns an error. This is very intentional and is particularly useful with load-balanced setups as in this case the server has an option to tell the health-check-monitoring loadbalancer if they are incapable of receiving traffic. The `Conn.handle` was also modified to be in conform with RFC5321 regarding proper handling of rejected sessions.

## Checkmarks

- [x] implementation
- [x] tests
- [ ] documentation: not found anything besides godoc.

### DIFF
--- a/server.go
+++ b/server.go
@@ -40,6 +40,7 @@ type Server struct {
 	ErrorLog          Logger
 	ReadTimeout       time.Duration
 	WriteTimeout      time.Duration
+	domainFn          func(*Conn) (string, error)
 
 	// Advertise SMTPUTF8 (RFC 6531) capability.
 	// Should be used only if backend supports it.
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -148,6 +149,19 @@ func (s *Server) Serve(l net.Listener) error {
 	}
 }
 
+// Getter for the optional dynamic domain string generator function
+func (s *Server) GetDomain(c *Conn) (string, error) {
```
emersion, line 153: This would better be implemented at the backend level.

endreszabo: if so, how? Initial greeting banner with the domain is presented by `Conn.greet()` called by the `server` that has just `Accept()`ed the connection and the backend is not even initialized before client issues an EHLO/HELO/LHLO in `Conn.handleGreet()`. I miss the point where the backend could have any control of this in its lifetime. Please advise.

PR #295, opened 2025-09-20
Add support for XCLIENT on server and client side

dejanstrbac:
Adding support for XCLIENT extension useful for proxying

https://www.postfix.org/XCLIENT_README.html

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -100,3 +100,13 @@ type AuthSession interface {
 	AuthMechanisms() []string
 	Auth(mech string) (sasl.Server, error)
 }
+
+// XCLIENTBackend is an add-on interface for Session. It provides support for the
+// XCLIENT extension.
+type XCLIENTBackend interface {
+	// XCLIENT handles the XCLIENT command. The attrs parameter contains
+	// the connection attributes provided by the client (ADDR, PORT, PROTO, 
+	// HELO, LOGIN, NAME). Backends can validate and process these attributes.
+	// If an error is returned, the XCLIENT command will be rejected.
+	XCLIENT(session Session, attrs map[string]string) error
+}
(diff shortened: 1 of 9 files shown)


endreszabo:
As a side note, I use the industry-standard [haproxy's Proxy Protocol](https://www.haproxy.com/blog/use-the-proxy-protocol-to-preserve-a-clients-ip-address) to preserve the client's IP (and more) with `go-smtp`. 
Implementing support for the Proxy Protocol needs no change to `go-smtp` codebase and is very easy to roll out:

```go
proxyListener := &proxyproto.Listener{
        Listener:   l,
}
defer proxyListener.Close()

if err := s.Serve(proxyListener); err != nil {
        log.Fatal(err)
}
```

dejanstrbac:
> As a side note, I use the industry-standard [haproxy's Proxy Protocol](https://www.haproxy.com/blog/use-the-proxy-protocol-to-preserve-a-clients-ip-address) to preserve the client's IP (and more) with `go-smtp`. Implementing support for the Proxy Protocol needs no change to `go-smtp` codebase and is very easy to roll out:
> 
> ```go
> proxyListener := &proxyproto.Listener{
>         Listener:   l,
> }
> defer proxyListener.Close()
> 
> if err := s.Serve(proxyListener); err != nil {
>         log.Fatal(err)
> }
> ```

I am aware of PROXY protocol, but Dovecot Proxy for LMTP delivery relies on XCLIENT extension which is our use case

endreszabo:
> but Dovecot Proxy for LMTP delivery

aah I see. I also use dovecot and was not aware of the existence of its LMPT server for delivery. Thus, I made a daemon that listens for gRPC calls from `go-smtp` and delivers the mails via the `dovelot-lda`. Not sure which one is better at this point :)

What would you say, what are the benefits for XCLIENT support with LMTP? Logging the original IP address?

MarkNLD:
Watching this! Thanks for the work.

PR #299, opened 2026-05-19
client: Autodetect need for SMTPUTF8, and use it

arnt:
This has no API changes; if you call SendMail() with some addresses that weren't supported before, then it'll either work (because the server supports the addresses) or break as before (because the server doesn't).


arnt:
Hi,

review, please? I need this code in something else…

emersion (approved):
Thanks for the ping! This patch LGTM.

PR #300, opened 2026-06-13
Handle error via errors.Is and remove unused code

aleksandr4842:


server.go
```diff
@@ -189,7 +189,7 @@ func (s *Server) handleConn(c *Conn) error {
 			if err == io.EOF || errors.Is(err, net.ErrClosed) {
 				return nil
 			}
-			if err == ErrTooLongLine {
+			if errors.Is(err, ErrTooLongLine) {
```
emersion, line 192: What is the motivation for this change?

aleksandr4842: This MR is just cosmetic. Handle all errors via errors.Is

PR #302, opened 2026-07-22
client: treat SASL responses as command data

HNO3Miracle:
## Summary

Pass SASL authentication commands through the fixed `%s` format instead of treating the command itself as a format string. This avoids `fmt` interpreting percent signs in encoded or future authentication data.

## Testing

- `go test ./...`

Assisted-by: OpenAI Codex <noreply@openai.com>


emersion (approved):
Good catch, LGTM!

PR #303, opened 2026-07-23
server: allow advertising extra EHLO capabilities

christhomas:
This adds an `ExtraCaps` field to `Server` holding additional capability
lines to advertise in response to EHLO. Each entry is advertised verbatim
as a separate capability line, after the built-in capabilities:

    srv := smtp.NewServer(be)
    srv.ExtraCaps = []string{"XEXAMPLE https://example.org/smtp-ext"}

    C: EHLO client.example.org
    S: 250-mx.example.org Hello client.example.org
    S: ...
    S: 250 XEXAMPLE https://example.org/smtp-ext

Motivation: servers sometimes need to advertise site-specific extensions
that go-smtp cannot know about — in our case a discovery capability that
points peer servers at an HTTPS upgrade endpoint (`RESTMAIL
https://<host>/restmail`). Today the capability list is built entirely
inside `handleGreet` from hard-wired `Enable*` flags, so the only options
are forking or proxying the connection. A verbatim string slice is the
smallest hook that covers this: no callback, no per-connection state, and
zero overhead for servers that do not set it.

Handling any commands such an extension defines remains the caller's
responsibility (same contract as the existing `Enable*` flags, which only
control advertising).

Includes a test (`TestServerExtraCaps`) in the style of the existing
capability tests via `testServerEhlo`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -308,6 +308,7 @@ func (c *Conn) handleGreet(enhanced bool, arg string) {
 			caps = append(caps, fmt.Sprintf("MT-PRIORITY %s", c.server.MtPriorityProfile))
 		}
 	}
+	caps = append(caps, c.server.ExtraCaps...)
 
 	args := []string{"Hello " + domain}
 	args = append(args, caps...)
--- a/server.go
+++ b/server.go
@@ -79,6 +79,15 @@ type Server struct {
 	// Default value of NONE to advertise no specific profile.
 	MtPriorityProfile PriorityProfile
 
+	// Additional capabilities to advertise in response to EHLO, e.g.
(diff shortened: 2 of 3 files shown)


emersion:
Sorry, but I have no interest in supporting non-standard SMTP extensions.

endreszabo:
well, this is not an extension that needs to be supported in the classic sense. this is a simple capability to add capability strings with minimal code change.

christhomas:
I mean, you're hardcoding the capability strings where this does not need to happen and SMTP is supposed to support extensions whether you like it or not. That is the stated function of this. So this change costs you absolutely nothing, just makes a hardcoded part of the code flexible and that makes it more valuable.

PR #304, opened 2026-07-27
chore: refresh github-guard hooks

christhomas:
Picks up the guard version that reads `.githooks/required-checks`, so this repo can state which status checks its default branch requires instead of having every discovered job name unioned into the required set.

That matters because a required check which no longer reports — a renamed job, a job refactored into a matrix, or a workflow whose `paths:` filters keep it from starting — reads as *pending* forever, and `enforce_admins` means an admin cannot merge past it either.

**No behaviour change here.** Without a `required-checks` file the guard discovers checks exactly as before. This is the payload refresh; declaring a gate is a separate, optional step.

Also carries an installer fix: re-installing no longer marks a repo's own files executable.

- antimatter-studios/agent-skills#26 — the declaration, plus that repo's first CI and tests
- antimatter-studios/agent-skills#27 — the installer fix

### DIFF
--- a/client.go
+++ b/client.go
@@ -380,7 +380,7 @@ func (c *Client) Auth(a sasl.Client) error {
 	} else if resp != nil {
 		resp64 = []byte{'='}
 	}
-	code, msg64, err := c.cmd(0, strings.TrimSpace(fmt.Sprintf("AUTH %s %s", mech, resp64)))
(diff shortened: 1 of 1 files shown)


christhomas:
Closing — opened against the wrong repository by mistake. This is a change to a fork's local git hooks and has nothing to do with go-smtp itself. Apologies for the noise.

PR #306, opened 2026-08-12
server: restore line limit when a BDAT transaction is reset

NotAFlightRisk:
Hit this on a server with CHUNKING on. If a client starts a BDAT transfer and then gives up with `RSET`, `MaxLineLength` never gets enforced again for the rest of that connection. Same story when a chunk trips `MaxMessageBytes`. Both of those end the transaction via `reset()`, which did not put the limit back.

Server left on the default `MaxLineLength` of 2000:

```
C: BDAT 8
C: Hey <3
S: 250 2.0.0 Continue
C: RSET
S: 250 2.0.0 Session reset
C: NOOP xxxxx... (2005 bytes)
S: 250 2.0.0 I have successfully done nothing
```

On a fresh connection that last line gets `500 5.4.0 Too long line, closing connection`.

The `io.Copy` failure path picked up a restore back in #139, these two go out through `reset()` instead. I have left both existing restores in `handleBdat` alone, the one before the LAST chunk still covers the `errPanic` return that skips `reset()`.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1341,6 +1341,9 @@ func (c *Conn) reset() {
 	c.bdatStatus = nil
 	c.bytesReceived = 0
 
+	// BDAT disables the line limit while reading a chunk, the transaction is over now.
+	c.lineLimitReader.LineLimit = c.server.MaxLineLength
+
 	if c.session != nil {
 		c.session.Reset()
 	}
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks!

PR #307, opened 2026-08-14
client: defer large SASL initial response instead of inlining it (#301)

### LINKED ISSUE #301: `Client.Auth` always inlines the SASL initial response, breaking mechanisms with a large initial response
`Client.Auth` sends the SASL initial response inline on the `AUTH <mech> <base64-ir>` command line. For mechanisms whose initial response is large — e.g. GSSAPI (RFC 4752), whose AP-REQ can be 1–2 KB, especially carrying a PAC — the command line exceeds SMTP's command-line limit (RFC 5321 §4.5.3.1.4: 512 octets). Strict servers reject it: Dovecot's submission service returns `500 5.5.2 Invalid command syntax` ("Line too long") before authentication.

RFC 4954 §4 allows the client to send `AUTH <mech>` without the initial-response argument and supply it in the first continuation instead. `Client.Auth` always inlines it and provides no way to defer it.

Reproduction: `Client.Auth` with a SASL GSSAPI client against Dovecot submission (`protocols = imap submission`, `auth_mechanisms = gssapi`) fails with `500 Line too long` before the AUTH exchange.

Expected: when the initial response is large, send `AUTH <mech>` and provide the initial response in the continuation, per RFC 4954 §4.

Note: continuation lines also have length limits (RFC 5321 §4.5.3.1.6: ~1000 octets), so very large tokens may still not fit; but the 512-octet command-line limit is hit first, and continuation-based initial response is the RFC-sanctioned behavior.

youdie006:
Fixes #301.

### Problem

`Client.Auth` always appends the base64 initial response to the `AUTH` command line:

```go
code, msg64, err := c.cmd(0, "%s", strings.TrimSpace(fmt.Sprintf("AUTH %s %s", mech, resp64)))
```

For SASL mechanisms whose initial response is large (e.g. GSSAPI, whose first token can be 1-2 KB with a PAC), that line exceeds the 512-octet command-line limit of RFC 5321 section 4.5.3.1.4. Strict servers (e.g. Dovecot submission) then reject it with `500 5.5.2 Line too long` before authentication can proceed.

### Fix

When inlining the initial response would overflow the 512-octet limit, send `AUTH <mech>` alone and provide the initial response as the reply to the first server challenge instead, as permitted by RFC 4954 section 4. Mechanisms whose initial response fits (the common case) are unchanged and still inline it.

### Test

Added `TestClientAuthLargeInitialResponse`: a SASL client returning a 600-byte initial response, driven through the existing `faker` mock. Red-green verified with `go test -run TestClientAuth ./...`: before the change the AUTH command line is 813 octets (initial response inlined, over the 512 limit); after, the command line is `AUTH LARGE` and the response is sent on its own line following the challenge. `gofmt` and `go vet` are clean.

---
Disclosure: I used AI assistance (Claude) while preparing this change. I verified the root cause against the RFCs, ran the tests (red-green), and take responsibility for the contribution.


emersion (commented):
Code structure is pretty hard to follow here. Can we send the too-large initial response outside of the loop?

youdie006:
You're right, thanks — restructured so the deferred response is sent before the loop, and the loop is byte-for-byte what it was before this PR.

```go
authCmd := strings.TrimSpace(fmt.Sprintf("AUTH %s %s", mech, resp64))
if len(authCmd)+2 > 512 && len(resp64) > 0 {
	// The initial response does not fit in the 512-octet command line (RFC 5321
	// section 4.5.3.1.4), so send it as the reply to the first challenge instead
	// (RFC 4954 section 4).
	code, msg64, err = c.cmd(0, "AUTH %s", mech)
	if err == nil && code == 334 {
		code, msg64, err = c.cmd(0, "%s", resp64)
	}
} else {
	code, msg64, err = c.cmd(0, "%s", authCmd)
}
for err == nil {
	// unchanged
```

The `deferredResp64` variable and the `continue` are gone. If the server answers the bare `AUTH <mech>` with something other than 334, that code falls into the loop and is handled by the existing `default` branch exactly as before, so the error paths are unchanged too.

Net change to `client.go` is +15/-1 against the merge base. `go test ./...` passes, `go vet` and `gofmt -l` clean. Reverting `client.go` and keeping the test still fails as it should:

```
--- FAIL: TestClientAuthLargeInitialResponse
    client_test.go:93: AUTH command line is 813 octets, exceeding the 512-octet limit
    client_test.go:98: deferred initial response was not sent
```

ml1nk:
My codebase differs a little bit, but I think this should work here too. In my understanding TrimSpace was added to remove the whitespace between %s and %s if resp64 is empty.

So if we reuse the branch we can remove the TrimSpace and extra fmt.Sprintf like this (https://github.com/uponusolutions/go-smtp/commit/21f3d248d03b3a118723a5f2270fab2464624ae6):
~~~
	// 512 - len("AUTH") - 2 * len(" ") - len(\r\n)
	if len(mech)+len(resp64) > 504 || len(resp64) == 0 {
		// The initial response (if any) does not fit in the 512-octet command line (RFC 5321
		// section 4.5.3.1.4), so send it as the reply to the first challenge instead
		// (RFC 4954 section 4). https://github.com/emersion/go-smtp/issues/301
		status, err = c.cmd(0, "AUTH %s", mech)
		if err == nil && status.Code == 334 && len(resp64) > 0 {
			status, err = c.cmd(0, "%s", resp64)
		}
	} else {
		status, err = c.cmd(0, "AUTH %s %s", mech, resp64)
	}
~~~

PR #308, opened 2026-08-14
client: clear recipient list when a new transaction starts

NotAFlightRisk:
Sending a second message down the same LMTP connection just hangs. `CloseWithLMTPResponse` reads one reply per entry in `c.rcpts`, and nothing clears that slice when a new transaction starts, so the second message waits on two replies, gets one, then sits there untill `SubmissionTimeout` - 12 minutes by default. The one reply it does get is filed against the wrong recipient.

`Reset` does clear it, but it's documented as aborting the current transaction, so you wouldn't reach for it after a delivery that went fine.

Only the LMTP path reads the field so plain SMTP shoud be unaffected. I've run it against this repo's own server and the scripted test, not a real delivery agent.

### DIFF
--- a/client.go
+++ b/client.go
@@ -34,7 +34,7 @@ type Client struct {
 	greetError error             // the error from the greeting
 	didHello   bool              // whether we've said HELO/EHLO/LHLO
 	helloError error             // the error from the hello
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks!

PR #309, opened 2026-08-18
server: consume the rest of an oversized message body in LMTP mode

NotAFlightRisk:
An LMTP server with `MaxMessageBytes` set stops reading the message at the limit and never picks up the rest. The command loop then carries on from wherever the reader stopped, and answers what's left of the body as commands.

With `MaxMessageBytes = 50` and a body whose third line happens to read `MAIL FROM:<root@example.org>`, a single `DATA` gets four replies:

```
552 5.3.4 <root@gchq.gov.uk> Maximum message size exceeded
500 5.5.2 Syntax errors,  CAN command unrecognized
250 2.0.0 Roger, accepting mail from <root@example.org>
501 5.5.2 Bad command
```

`CAN` is the tail of teh line the reader was cut off in the middle of, and the `501` is the terminating dot. From there on the client is three replies behind.

Plain `DATA` doesn't do this beacuse it drops the limit first. The LMTP handler has the same "make sure all the data has been consumed" copy in both of its branches, and neither one got the accompanying line.

Only bites with `MaxMessageBytes` non-zero. I've run it against this repo's test backend rather than a real delivery agent, so I dont know how one of those copes when its replies are out of step.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1215,6 +1215,7 @@ func (c *Conn) handleDataLMTP() {
 	if !ok {
 		// Fallback to using a single status for all recipients.
 		err := c.Session().Data(r)
+		r.limited = false
 		io.Copy(ioutil.Discard, r) // Make sure all the data has been consumed
 		for _, rcpt := range c.recipients {
 			status.SetStatus(rcpt, err)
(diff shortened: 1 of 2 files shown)


emersion (approved):
Thanks!

PR #310, opened 2026-08-25
docs: add navigable Sourcey API reference

paolosca90:
Adds a searchable, navigable API reference generated directly from the Go source, alongside the existing pkg.go.dev link. The live reference is at https://paolosca90.github.io/go-smtp/ and currently tracks commit 8d5af0d9db3ace5e4fdc0e8b427f5b157b0c6c44. The Sourcey config is included so the project can reproduce or adopt the build with npx sourcey build; generated HTML is not committed to the source branch. The site covers the root smtp package and cmd/smtp-debug-server, including functions, types, methods, examples, search, and llms.txt. Disclosure: this automation-assisted contribution was prepared while working Frantic bounty #113; the operator reviewed the generated output and is responsible for the proposal.


emersion:
Sorry, I'm not interested in this, GoDoc is more than enough.

PR #314, opened 2026-09-10
Move DELIVERBY and MT-PRIORITY to mail from #311

### LINKED ISSUE #311: MT-PRIORITY and DELIVERBY should be part of MAIL FROM and not RCPT TO
Currently MT-PRIORITY and DELIVERBY  are parameters in RCPT TO but the rfc states that they are part of MAIL FROM.

https://www.rfc-editor.org/info/rfc6710/

~~~
C: MAIL FROM:<eljefe@example.com> BY=120;R ENVID=QQ314159
            MT-PRIORITY=3
~~~

ml1nk:
This PR moves DELIVERBY and MT-PRIORITY from "rcpt to"  to "mail from" and improve range checks on the time argument. 

Closes #311

### DIFF
--- a/client.go
+++ b/client.go
@@ -477,6 +477,22 @@ func (c *Client) Mail(from string, opts *MailOptions) error {
 		}
 		// We can safely discard parameter if server does not support AUTH.
 	}
+	if _, ok := c.ext["DELIVERBY"]; ok && opts != nil && opts.DeliverBy != nil {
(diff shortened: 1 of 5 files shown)


emersion:
LGTM! Can you drop the "fix:" prefix, and use "client:" or "server:" instead? We don't use Conventional Commits.

ml1nk:
Like this?

emersion:
Thanks!

